PR #64, opened 2021-02-10
Update on task assignment

Jenniferstrej:
It was decided in the WG sync meeting that it would be easier for later contributors to volunteer if they saw tasks assigned. The chairs will take the unassigned tasks otherwise.


thschue (approved):
lgtm

OmerKahani (approved):

PR #69, opened 2021-02-14
Introduction to kubebuilder

feloy:
References https://github.com/cncf/sig-app-delivery/issues/65

### DIFF
--- a/operator-wg/whitepaper/051_kubebuilder.md
+++ b/operator-wg/whitepaper/051_kubebuilder.md
@@ -2,7 +2,7 @@
 
 The kubebuilder framework provides developers the possibilities to extend the Kubernetes API by using Custom Resource Definitions, and to create operators that handle these custom resources.
 
-The main concept provided by the kubebuilder framework is a *Manager*. In the same way the native Kubernetes controllers are grouped into a single Kubernetes Controller Manager (`kube-controller-manager`), you will be able to create several operators and make them managed by a single manager.
+The main entry point provided by the kubebuilder framework is a *Manager*. In the same way the native Kubernetes controllers are grouped into a single Kubernetes Controller Manager (`kube-controller-manager`), you will be able to create several contrllers and make them managed by a single manager.
 
 As Kubernetes API resources are attached to domains and arranged in Groups, Versions and Kinds, the Kubernetes custom resources you will define will be attached to your own domain, and arranged in your own groups, versions and kinds.
 
--- a/operator-wg/whitepaper/051_kubebuilder.md
+++ b/operator-wg/whitepaper/051_kubebuilder.md
@@ -1,0 +1,17 @@
+## kubebuilder Operator Framework
+
+The kubebuilder framework provides developers the possibilities to extend the Kubernetes API by using Custom Resource Definitions, and to create operators that handle these custom resources.
+
+The main concept provided by the kubebuilder framework is a *Manager*. In the same way the native Kubernetes controllers are grouped into a single Kubernetes Controller Manager (`kube-controller-manager`), you will be able to create several operators and make them managed by a single manager.
+
+As Kubernetes API resources are attached to domains and arranged in Groups, Versions and Kinds, the Kubernetes custom resources you will define will be attached to your own domain, and arranged in your own groups, versions and kinds.
+
+The first step when using kubebuilder is to create a project, that  will create the source code for building a single Manager attached to your domain.
+
+After initiating your project with a specific domain, you can add APIs to your domain and make these APIs managed by the manager.
+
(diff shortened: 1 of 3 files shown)


thschue (approved):
lgtm

operator-wg/whitepaper/051_kubebuilder.md
```diff
@@ -0,0 +1,17 @@
+## kubebuilder Operator Framework
+
+The kubebuilder framework provides developers the possibilities to extend the Kubernetes API by using Custom Resource Definitions, and to create operators that handle these custom resources.
+
+The main concept provided by the kubebuilder framework is a *Manager*. In the same way the native Kubernetes controllers are grouped into a single Kubernetes Controller Manager (`kube-controller-manager`), you will be able to create several operators and make them managed by a single manager.
```
coderanger, line 5: Disagree that this is the "main concept". It's certainly one of them but the main twp concepts are "apis" and "controllers".

Adirio: Have to agree with @coderanger.

operator-wg/whitepaper/051_kubebuilder.md
```diff
@@ -0,0 +1,17 @@
...
+
+As Kubernetes API resources are attached to domains and arranged in Groups, Versions and Kinds, the Kubernetes custom resources you will define will be attached to your own domain, and arranged in your own groups, versions and kinds.
+
+The first step when using kubebuilder is to create a project, that  will create the source code for building a single Manager attached to your domain.
```
coderanger, line 9: Also a single manager can handle any number of API groups on any number of domains.

operator-wg/whitepaper/051_kubebuilder.md
```diff
@@ -0,0 +1,17 @@
...
+
+After initiating your project with a specific domain, you can add APIs to your domain and make these APIs managed by the manager.
+
+Adding a resource to the project will generate some sample code for you: a sample *Custom Resource Definition* that you will adapt to build your own custom resource, and a sample *Reconciler* that will implement the reconcile loop for your operator handling this resource.
+
+The kubebuilder framework leverages the `controller-runtime` library, that provides the Manager and Reconciler concepts, among others.
```
coderanger, line 15: Should also mention controller-tools which provides the code gen.

operator-wg/whitepaper/051_kubebuilder.md
pwittrock: Main entry point?

Adirio: It is the main entry point but I don't know how relevant is this fact for a short description of the project. Maybe if you change it to this so that the "controller" concept is also present:
```suggestion
The main entry point provided by the kubebuilder framework is a *Manager*. In the same way the native Kubernetes controllers are grouped into a single Kubernetes Controller Manager (`kube-controller-manager`), you will be able to create several contrllers and make them managed by a single manager.
```

operator-wg/whitepaper/051_kubebuilder.md
Adirio: While technically speaking controller-tools is part of the kubebuilder ecosystem, its relevance to the end user is lower as it is abstracted/hidden behind a make target.

Adirio:
Kubebuilder used to be a scaffolder for operators in Go, but has evolved into a library for operators scaffolders that can be customized through plugins, and offers a plugin for Go operators. The information provided in this PR mainly relates to the kubebuilder Go plugin.

operator-wg/whitepaper/051_kubebuilder.md
feloy: Is it the normal behaviour of Kubebuilder, is it possible to handle several domains by using the CLI?

Adirio: Technically speaking, right now `""`, `"k8s.io"` and the domain you specify during init can be handled in the same project just by using the CLI. Additionally, kubebuilder is a scaffolder, meaning that it generates code but you have to edit and implement some parts, and changing the domain to a CRD is pretty easy and strightforward.

Outside of that, I think that what @coderanger means is that the manager is in no way bound to a single domain and this phrase is missleading in that sense.

feloy: I agree that the manager is able to handle several domains, but this phrase, and more generally this section, are describing kubebuilder.
It can be confusing if the reader understands that kubebuilder gives the possibility to handle several domains, but no options are found at the kubebuilder level to do this.

Adirio: Maybe?
```suggestion
The first step when using kubebuilder is to create a project attached to your domain, that  will create the source code for building a single Manager.
```

PR #114, opened 2021-06-01
Public Comment Version

thschue:
Signed-off-by: Thomas Schuetz <thomas.schuetz@dynatrace.com>


jlk:
The CNCF Supply Chain Security whitepaper has come out since the initial draft of the security chapter. Would like to add a reference to https://github.com/cncf/tag-security/tree/main/supply-chain-security/supply-chain-security-paper around line 421 in operator-wg/whitepaper/Operator-WhitePaper_v1-0.md.

bwplotka (commented):
This is amazing and thanks to PR based commenting phase, it's easy to give feedback. 

Overall I love this paper, I have a couple of small-ish suggestions, if you don't mind, based on our Red Hat experience:

1. When I read those two Operator Characteristics: Automation and Domain knowledge, those are really the same to me. Both apply automation thanks to domain knowledge, no?

>  #### Operational Automation
> #### Domain Knowledge

1. I think there is another very common Operator Characteristics or motivation for creating one. On top of Dyn. Configuration, Domain Knowledge Automation we can also see "Continous Provisioning/Installation" cases.

Simply put there is a certain pattern e.g used by [OLM](https://olm.operatorframework.io/) where the operator replaces the need of GitOps or overall CI/CD. Even Prometheus Operator is somehow doing it, by translating Prometheus CR into Prometheus yaml and applying it (nothing more magic that this inside).

Do you think it's worth covering?

1. Furthermore I wonder if there is any stance on merging those functionalities/characteristics together (aka "Monolith Operator"). I am really a fan of making sure things have a single purpose instead of "EverythingOperator". It's common to see multi-micro service application (e.g Thanos or Cortex) and operator deploying all of the components (despite all but one are stateless).

See: https://twitter.com/bwplotka/status/1395137433012998150

If that's not a good pattern, wonder if we can add section anti-patterns.


1. While we mention https://operatorhub.io/ it's maybe worth mentioning OLM which is behind the operator hub (https://olm.operatorframework.io/)?

1. I don't know if you are keen to mention it but we use https://github.com/brancz/locutus which offers nice abstraction over operator principles allowing us to create operator without touching any KubeAPI specific code. E.g we use it in https://github.com/observatorium/operator

BTW I am sneaking here thanks to @AlexsJones: https://twitter.com/AlexJonesax/status/1399777391719878662 🙃

operator-wg/whitepaper/Operator-WhitePaper_v1-0.md
```diff
@@ -109,9 +107,9 @@ The Operator pattern consists of three components:
 * The application or infrastructure that we want to manage.
 * A domain specific language that enables the user to specify the desired state of the application in a declarative way.
 * A controller that runs continuously:
-  * Reads and is aware of the state.
-  * Runs actions against the application in an automated way.
-  * Report the state of the application  in a declarative way.
+    * Reads and is aware of the state.
+    * Runs actions against the application in an automated way.
```
hongchaodeng, line 111: It's not always "against the application". It could be any operations, like cleaning up S3 buckets.
Maybe rephrase like: "Runs actions against change of operations state in an automated way".

hongchaodeng:
Thanks for the wonderful work and raising this PR for the community to jointly discuss. Operator pattern is now the most popular way for users to adopt and extend Kubernetes as their infrastructure platform.

One thing I want to mention or maybe add to the white paper is: as the number of Operators & CRDs adds up, its complexity of management also increases. For example, how to manage the conflicts between Operators, like two ingress-related functions? How to manage the dependencies and/or correlation of data flow between CRDs, like DB cluster and DB backup CRDs?

To resolve this problem, we would need some new models to manage Operators & CRDs and a new mechanism to oversee them with policy-base engine. When I originally participated in the design of [OLM](https://github.com/operator-framework/operator-lifecycle-manager) in CoreOS, we had wanted to solve this problem. We have continued this effort and provided more concrete model via [OAM](https://oam.dev/). A few more community efforts have also started, like this project called [kubeplus](https://github.com/cloud-ark/kubeplus).

This is a future look of what a k8s world full of Operators would look like. I think it is worth discussing and noting in the white paper.

bwplotka (commented):
~Something that we also often do, but I can't find any information about it here is hierarchical operators that "communicate" with other operators using it's CRs e.g:

Jaeger operator creating ElasticSearch CR for ES operator to provision it.
Some operator creating Service/Pod Monitor CR if they detect Prometheus Operator in the cluster.~

Ignore me, it's covered by stack creation (:

umangachapagain (commented):
`Writing operators that represent a whole stack` talks about a higher level operator that holds configuration for the entire stack but delegates the actual work to be done to individual component operators. And, `Best Practices` suggests that an operator shouldn't install other operators.

So, how do we go about installing these individual component operators?

umangachapagain (commented):
Controller vs Operator section seems unclear to me.

```
If the controller has additional operational knowledge like how to upgrade or remediate from errors, it is an operator.
```

When we say "additional operational knowledge", how do we define "additional"?
I think `deployment-controller` does some amount of error remediation and also handles upgrades, no?

dmesser:
> `Writing operators that represent a whole stack` talks about a higher level operator that holds configuration for the entire stack but delegates the actual work to be done to individual component operators. And, `Best Practices` suggests that an operator shouldn't install other operators.
> 
> So, how do we go about installing these individual component operators?

There are multiple ways to do this. An Operator distribution package could consist of multiple separate controllers, each handling their own part of the stack. Deploying the Operator would result in all controllers running at once but only a higher level API/CRD is actually exposed to end users. The controller responsible for this API (the higher level operator) delegates several duties to the other controllers that are part of it's packaged using "internal" CRDs. This is useful when the whole "stack" is owned and controlled by the Operator authors and the "subordinate" controllers don't make sense as a standalone Operator project. This is simply separation of concerns.

The other pattern is that Operators depend on other general-purpose Operator projects, to do to their work. You could image that an Operator might depend on cert-manager, the prometheus operator and the Crunchy PostgreSQL operator  to deploy its workload with rotating certificates, monitoring and a SQL database. In this case the higher level workload Operator should not try to ship and install cert-manager et al. This is because the Operator author then signs up for shipping and maintaining their versions as well and dealing with the general  problem area of CRD lifecycle management.
Instead a package management solution should be employed that supports dependency resolution at runtime, so that installing the other required Operators is delegated to a package manager and not part of the higher level Operator code. This is useful for Operators that depend on other Operators, which are useful on their own and might even be shared with multiple other Operators on the cluster. [OLM](https://github.com/operator-framework/operator-lifecycle-manager), part of the Operator Framework Project, is such a package manager.

danielpacak:
First off, thank you for writing this white paper. I’ve been looking for such a reference document for a long while.
On top of what’s been described I’d love to see the following topics covered / commented on by subject matter experts:
1. **HA and Scaling**: Scaling an operator itself (in HA environments). The document describes that an operator might scale the subordinate application but what if we want to scale an operator itself. IIRC we can scale up operator’s pods, but only one of them is supposed to be active by leveraging a leader election algorithm and lease objects (aka active-passive mode). As an example, imagine that you are writing an operator that discovers K8s workloads to perform long running / compute intensive tasks for each workload. Let’s assume that we want to run the tasks in parallel to speed up the reconciliation process. How would you do that with the operator pattern? The active-passive mode is only for redundancy.
2. **Designing Operators / Scaling**: Should each operator implement a proprietary scaling methods for the application that it’s managing or should we strive for a design that may leverage generic purpose scaling operators such as KEDA (Kubernetes Event Driven Autoscaling)
3. **Security**: You mentioned security, RBAC and least privilege principle as crucial. From my experience, implementing minimal permissions declaratively is very challenging because you have to know the install mode (i.e. namespaced vs cluster-wide operators) to decide whether to create Role and RoleBinding vs ClusterRole and ClusterRoleBindings. A tricky case is also when an operator is supposed to watch resources in multiple namespace, but not the whole cluster. To some extent Operator Lifecycle Manager is managing that based on instal modes and the ClusterServiceVersion, which allow you to define permission and cluster permission templates, but it has its limitations. That said I’d love to see some more hands on advice for operator developers on how to implement the least privileged principle. Are there any RBAC permission templating solutions out there for upstream Kubernetes for those who cannot or don’t want to use OLM?
4. **Operator Catalogs**: You mentioned OperatorHub and ArtifactHUB, which both rely on a central catalog of operator manifests. Could you elaborate on that approach and describe what’s the advantage of such a catalog over a Helm chart or static YAML descriptors? I’ve heard many times from end users that OLM is overkill and Helm charts are usually buggy. Surprisingly end users often prefer static YAMLs, but with static YAML implementing least privileges is probably not possible. So I’m wondering what should be a standardized (answer) way to deploy operators.
5. **Development / Testing**: I’m always looking for good advice on testing software such as operators. On one hand, unit testing certain execution paths with API fakes and mocks seems to be appropriate. On the other hand, this does not give enough confidence to roll out a new version based on test outcome. What’s more, we can change API calls upside down in a given reconciler implementation, yet keep the behavior of the operator the same.  Can you share any advice on unit / integration / component / e2e / chaos testing and frameworks that might facilitate that, e.g. Ginkgo, Gomega?

umangachapagain:
> > `Writing operators that represent a whole stack` talks about a higher level operator that holds configuration for the entire stack but delegates the actual work to be done to individual component operators. And, `Best Practices` suggests that an operator shouldn't install other operators.
> > So, how do we go about installing these individual component operators?
> 
> There are multiple ways to do this.........[OLM](https://github.com/operator-framework/operator-lifecycle-manager), part of the Operator Framework Project, is such a package manager.
>
@dmesser Isn't OLM itself an operator that does package management?
It uses CRDs for custom configuration, it has reconciles, it does upgrades, uninstall etc.

I guess I don't understand where to draw line between controllers, operators and other automation softwares. It all seems the same.

dmesser:
> @dmesser Isn't OLM itself an operator that does package management?
> It uses CRDs for custom configuration, it has reconciles, it does upgrades, uninstall etc.
> 
> I guess I don't understand where to draw line between controllers, operators and other automation softwares. It all seems the same.

Yeah, it's one of those tendencies in modern software architecture - self-hosting (git hosting git source code, linux servers running kernel.org, Kubernetes hosting Kubernetes, etc). OLM was created on the core believe that CRD lifecycle and package management should happen on the cluster and constantly, not outside the cluster on a one-off basis. Hence choosing the Operator implementation route.

To respond to your question: controller is the implementation, Operator is the pattern of using custom controllers with CRDs and automation is what is looking to be achieved with this. They are all related. Hope this helps :)

dmesser:
> 1. **HA and Scaling**: Scaling an operator itself (in HA environments). The document describes that an operator might scale the subordinate application but what if we want to scale an operator itself. IIRC we can scale up operator’s pods, but only one of them is supposed to be active by leveraging a leader election algorithm and lease objects (aka active-passive mode). As an example, imagine that you are writing an operator that discovers K8s workloads to perform long running / compute intensive tasks for each workload. Let’s assume that we want to run the tasks in parallel to speed up the reconciliation process. How would you do that with the operator pattern? (Even if the active-passive mode is only for redundancy, I believe it's worth describing as well.)

I think it's certainly possible to do some form of sharding so that multiple controller instances would share the workload. But before you'd go there, I would ask the question why it is needed. Is it because the long running / compute intensive task is part of the reconciliation? If so, the solution to me seems to externalize it and hand it over to Kubernetes jobs, which state are reconciled in the controller but they wouldn't block the reconciltion process from progressing.

> 2. **Designing Operators / Scaling**: Should each operator implement a proprietary scaling methods for the application that it’s managing or should we strive for a design that may leverage generic purpose scaling operators such as KEDA (Kubernetes Event Driven Autoscaling)
> 3. **Designing Operators / Loop**: Frameworks such as controller-runtime allow building reconcilers that are triggered by events / changes to K8s objects. Is there any similar framework / pattern for triggering reconciliation based on time? Even if it's as simple as a ticker in Go I'd mention it as a viable hand crafted solution.

Similar to my first comment, if an Operator needs certain things to happen at a certain time, what would be wrong with the Operator creating a `CronJob`?

> 4. **Security**: You mentioned security, RBAC and least privilege principle as crucial. From my experience, implementing minimal permissions declaratively is very challenging because you have to know the install mode (i.e. namespaced vs cluster-wide operators) to decide whether to create Role and RoleBinding vs ClusterRole and ClusterRoleBindings. A tricky case is also when an operator is supposed to watch resources in multiple namespaces, but not the whole cluster. To some extent Operator Lifecycle Manager is managing that based on instal modes and the ClusterServiceVersion, which allow you to define permission and cluster permission templates, but it has its limitations. That said I’d love to see some more hands on advice for operator developers on how to implement the least privileged principle. Are there any RBAC permission templating solutions out there for upstream Kubernetes for those who cannot or don’t want to use OLM?

We are looking at things like https://github.com/FairwindsOps/rbac-manager for this, so we can delegate the RBAC management away from OLM to some extent. But that's just moving the problem elsewhere. In OLM we are looking at a future where all Operators are running at the cluster level, but there are distinct sets of RBAC that are given to tenants who are supposed to be able to use these Operators and another distinct set of RBAC that is given to the Operator. We plan to make the Operator developer be able to control what those sets look like. That should remove the limitations that OLM has today in terms of automating RBAC scope. WDYT?

> 5. **Operator Catalogs**: You mentioned OperatorHub and ArtifactHUB, which both rely on a central catalog of operator manifests. Could you elaborate on that approach and describe what’s the advantage of such a catalog over a Helm chart or static YAML descriptors? I’ve heard many times from end users that OLM is overkill and Helm charts are usually buggy. Surprisingly end users often prefer static YAMLs, but with static YAML implementing least privileges is probably not possible. So I’m wondering what should be a standardized (answer) way to deploy operators.

I would counter the argument that OLM is overkill with the assessment that today, Operator developers do not address important concerns like Operator discoverability, updates, dependency resolution, content distribution and CRD lifecycle management. This is more of an awareness issue rather than deliberate oversight. But long term it decides whether your Operator is fit for practical use.
I can relate to the impression that it feels overkill if one doesn't care about these things but users will definitely care about it. As an upstream developer a simple set of `kustomize` overlays to distribute, configure and install your Operator might be enough to get a contributor started. It is also probably enough for and end user if it's ever going to be the only Operator running on a cluster and if there is ever only one tenant. But a production cluster where multiple Operators with their own lifecycles and requirements are going to be running and you may not know or trust the tenants? I don't see a way around bespoke solution for this but there is definitely room for making it easier. You may want to look here https://github.com/operator-framework/enhancements/blob/ce43d4c44277ae70f99b0297a15d32ac57b27d08/enhancements/csvless-bundles.md

> 6. **Development / Testing**: I’m always looking for good advice on testing software such as operators. On one hand, unit testing certain execution paths with API fakes and mocks seems to be appropriate. On the other hand, this does not give enough confidence to roll out a new version based on test outcome. What’s more, we can change API calls upside down in a given reconciler implementation, yet keep the behavior of the operator the same.  Can you share any advice on unit / integration / component / e2e / chaos testing and frameworks that might facilitate that, e.g. Ginkgo, Gomega? In particular, how would you test in automated way that your operator migrated app from V1 to V2? How would you simulate / recover failure in automated test? Would it require changes to the app itself?

One approach we've looked at is kuttl to do k8s-level black box testing. We've implemented a fairly generic custom functional test framework on top of that in Operator SDK: https://sdk.operatorframework.io/docs/advanced-topics/scorecard/custom-tests/ This is more of a test harness or execution framework rather than a specific way to write the test logic. But it would address your automation question. Thoughts?

thschue:
> The CNCF Supply Chain Security whitepaper has come out since the initial draft of the security chapter. Would like to add a reference to https://github.com/cncf/tag-security/tree/main/supply-chain-security/supply-chain-security-paper around line 421 in operator-wg/whitepaper/Operator-WhitePaper_v1-0.md.

Is this ok for you? -> https://github.com/cncf/tag-app-delivery/pull/115

jlk:
> We are looking at things like https://github.com/FairwindsOps/rbac-manager for this, so we can delegate the RBAC management away from OLM to some extent. But that's just moving the problem elsewhere. In OLM we are looking at a future where all Operators are running at the cluster level, but there are distinct sets of RBAC that are given to tenants who are supposed to be able to use these Operators and another distinct set of RBAC that is given to the Operator. We plan to make the Operator developer be able to control what those sets look like. That should remove the limitations that OLM has today in terms of automating RBAC scope. WDYT?

This sounds interesting @dmesser. Would want to make sure those two parts are coded carefully - tenant breakout via cluster operator could be a Bad Thing...

I think there's use cases where some operators don't need full cluster access, but maybe that's handled outside of OLM?

Jenniferstrej:
Thanks a lot for all your comments @bwplotka! We had a discussion on all of your comments and have specific actions for one of them but to be more objective in the others we thought that it's better if you could suggest specific changes that we can then add without relying on our interpretation alone (also avoid too many back and forths and encourage more active contributions :)). Please see details below:

> This is amazing and thanks for the PR based commenting phase, it's so easy to give feedback that way.
> 
> Overall I love this paper, I have a couple of small-ish suggestions, if you don't mind, based on our Red Hat experience:
> 
> 1. When I read those two Operator Characteristics: Automation and Domain knowledge, those are really the same to me. Both apply automation thanks to domain knowledge, no?

Could you clarify your thoughts to us on a PR suggestion, please?

> 
> > #### Operational Automation
> > #### Domain Knowledge
> 
> 1. I think there is another very common Operator Characteristics or motivation for creating one. On top of Dyn. Configuration, Domain Knowledge Automation we can also see "Continous Provisioning/Installation" cases.

Could you suggest the cases you’d like to see here?

> 
> Simply put there is a certain pattern e.g used by [OLM](https://olm.operatorframework.io/) where the operator replaces the need of GitOps or overall CI/CD. Even Prometheus Operator is somehow doing it, by translating Prometheus CR into Prometheus yaml and applying it (nothing more magic that this inside).
> 
> Do you think it's worth covering?
> 
> 1. Furthermore I wonder if there is any stance on merging those functionalities/characteristics together (aka "Monolith Operator"). I am really a fan of making sure things have a single purpose instead of "EverythingOperator". It's common to see multi-micro service application (e.g Thanos or Cortex) and operator deploying all of the components (despite all but one are stateless).

Action: https://github.com/cncf/tag-app-delivery/issues/118

> 
> See: https://twitter.com/bwplotka/status/1395137433012998150
> 
> If that's not a good pattern, wonder if we can add section anti-patterns.
> 
> 1. While we mention https://operatorhub.io/ it's maybe worth mentioning OLM which is behind the operator hub (https://olm.operatorframework.io/)?
> 2. I don't know if you are keen to mention it but we use https://github.com/brancz/locutus which offers nice abstraction over operator principles allowing us to create operator without touching any KubeAPI specific code. E.g we use it in https://github.com/observatorium/operator

Please feel free to add a PR covering this. Thanks!

> 
> BTW I am sneaking here thanks to @AlexsJones: https://twitter.com/AlexJonesax/status/1399777391719878662 🙃

operator-wg/whitepaper/Operator-WhitePaper_v1-0.md
Jenniferstrej: Suggestion applied https://github.com/cncf/tag-app-delivery/pull/114/commits/713d692a5c3851de7e28173c9f26b655eb5619f0 - Thanks!

Jenniferstrej:
> Controller vs Operator section seems unclear to me.
> 
> ```
> If the controller has additional operational knowledge like how to upgrade or remediate from errors, it is an operator.
> ```
> 
> When we say "additional operational knowledge", how do we define "additional"?
> I think `deployment-controller` does some amount of error remediation and also handles upgrades, no?

We've removed the word "additional" to reduce confusion. Commit https://github.com/cncf/tag-app-delivery/pull/114/commits/b47c29648094a69c7ad699c88aa40ab81c7c2daa

Jenniferstrej:
> First off, thank you for writing this white paper. I’ve been looking for such a reference document for a long while.
> On top of what’s been described I’d love to see the following topics covered / commented on by subject matter experts:
> 
> 1. **HA and Scaling**: Scaling an operator itself (in HA environments). The document describes that an operator might scale the subordinate application but what if we want to scale an operator itself. IIRC we can scale up operator’s pods, but only one of them is supposed to be active by leveraging a leader election algorithm and lease objects (aka active-passive mode). As an example, imagine that you are writing an operator that discovers K8s workloads to perform long running / compute intensive tasks for each workload. Let’s assume that we want to run the tasks in parallel to speed up the reconciliation process. How would you do that with the operator pattern? (Even if the active-passive mode is only for redundancy, I believe it's worth describing as well.)
> 2. **Designing Operators / Scaling**: Should each operator implement a proprietary scaling methods for the application that it’s managing or should we strive for a design that may leverage generic purpose scaling operators such as KEDA (Kubernetes Event Driven Autoscaling)
> 3. **Designing Operators / Loop**: Frameworks such as controller-runtime allow building reconcilers that are triggered by events / changes to K8s objects. Is there any similar framework / pattern for triggering reconciliation based on time? Even if it's as simple as a ticker in Go I'd mention it as a viable hand crafted solution.
> 4. **Security**: You mentioned security, RBAC and least privilege principle as crucial. From my experience, implementing minimal permissions declaratively is very challenging because you have to know the install mode (i.e. namespaced vs cluster-wide operators) to decide whether to create Role and RoleBinding vs ClusterRole and ClusterRoleBindings. A tricky case is also when an operator is supposed to watch resources in multiple namespaces, but not the whole cluster. To some extent Operator Lifecycle Manager is managing that based on instal modes and the ClusterServiceVersion, which allow you to define permission and cluster permission templates, but it has its limitations. That said I’d love to see some more hands on advice for operator developers on how to implement the least privileged principle. Are there any RBAC permission templating solutions out there for upstream Kubernetes for those who cannot or don’t want to use OLM?
> 5. **Operator Catalogs**: You mentioned OperatorHub and ArtifactHUB, which both rely on a central catalog of operator manifests. Could you elaborate on that approach and describe what’s the advantage of such a catalog over a Helm chart or static YAML descriptors? I’ve heard many times from end users that OLM is overkill and Helm charts are usually buggy. Surprisingly end users often prefer static YAMLs, but with static YAML implementing least privileges is probably not possible. So I’m wondering what should be a standardized (answer) way to deploy operators.
> 6. **Development / Testing**: I’m always looking for good advice on testing software such as operators. On one hand, unit testing certain execution paths with API fakes and mocks seems to be appropriate. On the other hand, this does not give enough confidence to roll out a new version based on test outcome. What’s more, we can change API calls upside down in a given reconciler implementation, yet keep the behavior of the operator the same.  Can you share any advice on unit / integration / component / e2e / chaos testing and frameworks that might facilitate that, e.g. Ginkgo, Gomega? In particular, how would you test in automated way that your operator migrated app from V1 to V2? How would you simulate / recover failure in automated test? Would it require changes to the app itself?

We discussed this and would love to have contributions on these topics and will take them into account on a second version of the whitepaper if you can’t this time.

Jenniferstrej:
> > Writing operators that represent a whole stack talks about a higher level operator that holds configuration for the entire stack but delegates the actual work to be done to individual component operators. And, Best Practices suggests that an operator shouldn't install other operators.
> 
> > So, how do we go about installing these individual component operators?
> 
> There are multiple ways to do this. An Operator distribution package could consist of multiple separate controllers, each handling their own part of the stack. Deploying the Operator would result in all controllers running at once but only a higher level API/CRD is actually exposed to end users. The controller responsible for this API (the higher level operator) delegates several duties to the other controllers that are part of it's packaged using "internal" CRDs. This is useful when the whole "stack" is owned and controlled by the Operator authors and the "subordinate" controllers don't make sense as a standalone Operator project. This is simply separation of concerns.
> 
> The other pattern is that Operators depend on other general-purpose Operator projects, to do to their work. You could image that an Operator might depend on cert-manager, the prometheus operator and the Crunchy PostgreSQL operator to deploy its workload with rotating certificates, monitoring and a SQL database. In this case the higher level workload Operator should not try to ship and install cert-manager et al. This is because the Operator author then signs up for shipping and maintaining their versions as well and dealing with the general problem area of CRD lifecycle management.
> Instead a package management solution should be employed that supports dependency resolution at runtime, so that installing the other required Operators is delegated to a package manager and not part of the higher level Operator code. This is useful for Operators that depend on other Operators, which are useful on their own and might even be shared with multiple other Operators on the cluster. [OLM](https://github.com/operator-framework/operator-lifecycle-manager), part of the Operator Framework Project, is such a package manager.

Action: https://github.com/cncf/tag-app-delivery/projects/1#card-63259453 - cc @thschue

AlexsJones (approved):

OmerKahani (approved):

Jenniferstrej (approved):
🚀

AlexsJones (approved):

PR #179, opened 2022-01-18
[gitops-wg] Adding self to interested parties and adding github username column

rumstead:


scottrigby (approved):
@rumstead Hello! 😄 

And also thank you for this follow up to https://github.com/gitops-working-group/gitops-working-group/pull/170. If anyone is wondering why there's an additional column, this was a request.

PR #181, opened 2022-01-18
[gitops-wg] Moving meeting notes to GH

chris-short:
Meeting artifacts will live here and on YouTube moving

This is an old commit that wasn't merged before the main branch rename.


scottrigby (approved):
Thanks for re-making this @chris-short 💯

PR #202, opened 2022-02-16
Add organization size field to issue forms

nagyv:


joshgav (approved):
This would be useful info and a dropdown is easy for folks too. Thanks for the suggestion @nagyv.

AlexsJones (approved):

PR #232, opened 2022-06-04
Update GitOps WG logo to use CNCF URL

niklasmtj:
This PR updates the path to the GitOps WG Logo to use the CNCF URL.
My Markdown extension seems to also formatted the table for better markdown readability, please tell me if it's okay, otherwise I will change it back.


oleg-nenashev (approved):
Would be nice to avoid formatting changes though

scottrigby (approved):
Thanks for pinging, I missed this one! Nice work, the broken image had been annoying me but hadn't found time to fix 👏

PR #234, opened 2022-06-14
Add GitOps Days 2022

dholbach:
Signed-off-by: Daniel Holbach <daniel@weave.works>


oleg-nenashev (approved):
Better late than never

scottrigby (approved):
This is fine for now 👍 

Ultimately we want to de-duplicate these timeline items that are already on the gitops events list at https://OpenGitOps.dev/events. Might be good to link to that from here when we do ❤️

scottrigby (changes requested):
Actually @dholbach we need to resolve conflicts, otherwise looks good

scottrigby (approved):
👍

PR #235, opened 2022-06-15
Update CHAIRS.md

chris-short:
Creating an official open slot on July 1st as that is my last day being a co-chair.

Please do not merge until after July 1.

Signed-off-by: Chris Short \<cbshort@amazon.com>


chris-short:
Dag nabbit DCO

PR #250, opened 2022-11-07
fix: update organization on rs

roberthstrand:
As I have changed companies, I would like to update the interested parties list to reflect that.

Signed-off-by: Roberth Strand <me@robstr.dev>


joshgav (approved):
LGTM. Perhaps we should remove these docs though? What is their purpose?

roberthstrand:
> LGTM. Perhaps we should remove these docs though? What is their purpose?

I think that is a valid question, perhaps something to discuss in the next TAG meeting?

PR #237, opened 2022-07-27
added my name and company to interested-parties

antnsn:
added my name and company to interested-parties

PR #256, opened 2022-12-14
Remove "interested parties" doc

joshgav:
The interested parties doc in the GitOps WG directory was intended to show broad support for forming that group. Now that the group is well-established the doc doesn't serve any further purpose, but new contributors are confused and think they should add themselves, leading to extraneous PRs in this repo and people asking for similar docs for other groups.

In a recent TAG meeting we agreed to remove this doc and this is the PR for that. It closes #254, closes #250, closes #240, and closes #237.


roberthstrand (approved):
I agree, thanks for following up on this.

joshgav:
Woohoo, closed a bunch of PRs!

scottrigby:
Hm, this doc was part of the OpenGitOps governance processes: https://github.com/open-gitops/project/blob/main/GOVERNANCE.md

We will have to look into how to resolve this. A change in governance there may be in order. But the point is to allow actually invested/interested people vote, and protect a last minute mass voting sneak attack 😄 (you must have been on the list at least 1 week prior to the vote in order for your vote to count).

PR #257, opened 2022-12-14
[WG Platforms] Framework for platforms paper

joshgav:
This is a proposal for a framework to build and publish the "Platforms Definition" whitepaper we're working on in WG Platforms. I went ahead and included all the current content from [the doc in progress](https://docs.google.com/document/d/1UDL3E5BqPyDdzV1wek9o8xR-nLJQXEEMU1qIMCgq5cI/) as well.

It uses [bikeshed](https://github.com/tabatkins/bikeshed) to generate a complete HTML doc from a template and the section docs. You can try it by cloning this PR and running `bikeshed` within the paper's directory; then open the generated `index.html` in a browser to review.

bikeshed is actively maintained by W3C and used by them and TC39 already so it seems like a project CNCF can rely on and perhaps participate in.

This PR also includes and uses the [w3c/spec-prod](https://github.com/w3c/spec-prod) GitHub Action step to invoke bikeshed, generate the doc, and push it into the `jekyll` branch as a post. As a nice bonus, the generated doc is also available as an artifact after the Action runs for review. See an example in this run in my repo fork: <https://github.com/joshgav/tag-app-delivery/actions/runs/3699614235>. The action is currently configured to only be run on demand, not on PRs; we can adapt that as we go.

The generated doc still needs some refinement but I wanted to go ahead and share early. LMK what you think!

Issue #246, opened 2022-10-23
[Platforms] Publish definition of "platforms" for cloud computing


joshgav:
Hi folks - WG Platforms proposes to publish a doc on the meaning, value, attributes and capabilities of platforms to help enterprise leaders, enterprise architects and platform product managers determine the value of platforms and what to expect and require from them.

We'd like to create this doc in the [WG Platforms subdirectory](https://github.com/cncf/tag-app-delivery/tree/main/platforms-wg) and publish it on the TAG and CNCF blogs.

We expect that contributors will raise issues about the doc once it's in GitHub. We'll also run live interviews after initial publication to continuously update our positions. We'll try to get the initial version published by end of 2022 - that gives us ~70 days.

While this first iteration will be mostly based on the observations and analysis of experts and already-published experience reports, we'll follow up by gathering more experience reports from CNCF end users/enterprises to inform the next iteration, targeting summer 2023.

Work is underway in <https://docs.google.com/document/d/1UDL3E5BqPyDdzV1wek9o8xR-nLJQXEEMU1qIMCgq5cI/>. That doc is open to all to comment; and if you want to edit click "Share" and then "Ask to share". We'll discuss this live in WG Platforms meetings and asynchronously in the [WG Platforms Slack](https://cloud-native.slack.com/archives/C020RHD43BP) as well as this thread.

The report will discuss the following topics; but this list is just to jog our thinking. What other topics and sections should we include?

Thanks in advance for your contributions and collaboration!

## Proposed topics

- What is a _platform_ in relation to cloud-native computing?
    - Platform means many things in societies and systems. What are the key characteristics of a "cloud-native computing" platform? What is its purpose, who does it serve, and how can its impact be measured?
    - Who are platforms intended to serve? For example: application developers, ML model developers, COTS operators, information workers. Is there a common "base" platform or will several platforms be needed in big organizations?
    - What names and descriptors are appropriate for various platforms and what do they imply? Terms found in industry chatter include "developer platform", "developer experience", "cloud platform", "app platform", "delivery platform", "MLOps platform" and more.
- What value do platforms and platform engineering promise cloud-native computing practictioners and enterprises?
    - What are the reasons platforms and "platform engineering" have trended recently in the cloud-native community?
    - In what way (if at all) do they differ from [DevOps](https://www.google.com/search?q=devops+manifesto) and [Reliability Engineering](https://www.google.com/search?q=reliability+engineering)?
[...]

PR #258, opened 2022-12-21
[WG Platforms] Proposal for platform attributes

joshgav:
A proposal for platform attributes, part of #246 and #257.


joshgav:
Replaced by #265

PR #259, opened 2022-12-21
[WG Platforms] Proposal for platform core definition

joshgav:
A proposal for platform core definition, part of #246 and #257, see also #258


joshgav:
Replaced by #263

PR #265, opened 2023-01-03
[WG Platforms] Whitepaper - Platform attributes

joshgav:


platforms-whitepaper/v1alpha1/40-platform-attributes.md
```diff
@@ -1,3 +1,52 @@
 ## Attributes of platforms
 
-// TODO
\ No newline at end of file
+Following are key attributes that affect the success of a cloud or developer
```
ThomasVitale, line 3: Up to this point, we used the generic term "platform" rather than "cloud platform" or "developer platform". Should we do the same here? It might be leftovers from before we decided on how to name them.

`Following are key attributes that affect the success of a platform, which should:`

Issue #277, opened 2023-01-10
[WG Platforms] Platforms Def - Platform challenges


joshgav:
Review and finalize "Challenges of platforms" section of Platforms whitepaper: <https://github.com/cncf/tag-app-delivery/blob/platforms-v1alpha1/platforms-whitepaper/v1alpha1/paper.md#challenges-with-platforms>

joshgav:
Consider integrating challenges for platform teams in this section.

taras:
I'll support this

roberthstrand:
I can support as well

joshgav:
Thank you for #285 and other discussions!

Issue #276, opened 2023-01-10
[WG Platforms] Platforms Def - Platform team attributes


joshgav:
Review and finalize "Platform team attributes" section of Platforms whitepaper: <https://github.com/cncf/tag-app-delivery/blob/platforms-v1alpha1/platforms-whitepaper/v1alpha1/paper.md#attributes-of-platform-teams>

taras:
I would like to provide some input about DX teams

joshgav:
@roberthstrand did you want to be responsible for this with Taras? Assignees will gather edits and help finalize the content by 2/28.

roberthstrand:
Yes please, @joshgav 😄

joshgav:
Thank you all for #309, #305, #301, #294, #289 and more. The paper is much better after your contributions!

PR #285, opened 2023-01-18
Integrate platform team challenges into general challenges

lianmakesthings:
#276 #277
As discussed in the 01/17 WG platforms meeting, I have created a suggestion to move the "Enabling platform teams" section into "Challenges with platforms"

### DIFF
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -278,7 +278,7 @@ To counter this, platform teams should include product managers from the start
 to share roadmaps, gather feedback and generally understand and represent the
 needs of platform users.
 
-However, Platform teams can easily be overloaded in their efforts to support many
+Additionally, Platform teams can easily be overloaded in their efforts to support many
 product teams simultaneously. Ways to reduce load on the platform team include
 the following:
 
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -267,7 +267,7 @@ of engaged and skillful app teams. Detailed feedback from such teams improves th
 first platform experiences; and people from those teams help champion and
 evangelize the platform to later adopters.
 
-Perhaps most important is to recognize that the success of a platform is
+Perhaps most important is to treat the platform as a customer-facing product and recognize that its success is
 directly dependent on the success of its users and products; and as such it's
 vital that platform teams partner with app teams and other users to prioritize,
 plan, implement and iterate on the platform's capabilities and user experiences.
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -288,13 +288,18 @@ stream teams (see the previous two paragraphs), presenting the platform team as
 a strategic partner of product teams in delivering value to customers.
 
 ## Enabling platform teams
-Another challenge for platform teams is the amount of cognitive load, when attempting
-to support multiple teams simultaneously.
 
-Ways to reduce load on the platform team include the following:
+It is clear from these challenges that platform teams are faced with a number of
+diverse responsibilities which lead to cognitive load. Just as with their
+application team counterparts, this challenge grows with the number and diversity
+of users and teams they need to support.
+
+It is important to focus the platform team's energy on the experience and
+capabilities that are unique to their specific business. Ways to reduce load on
+the platform team include the following:
 
 1. Use implementations from managed service providers where reasonable
-1. Use open source frameworks and toolkits for creating docs, templates and
+1. Leverage open source frameworks and toolkits for creating docs, templates and
    compositions for application team use
 1. Ensure platform teams are staffed appropriately for their domain and number
    of customers
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -259,14 +259,6 @@ following which implementers should keep in mind.
 1. Platform teams must seek support of enterprise leadership and show impact on
    value streams
 
-When adopting platforms, choosing the right capabilities and experiences to
-enable first, can be crucial. Capabilities that are frequently required and
-undifferentiated, like pipelines, databases and observability, may be a good
-place to start. Platform teams may also choose to focus first on a limited number
-of engaged and skillful app teams. Detailed feedback from such teams improves the
-first platform experiences; and people from those teams help champion and
-evangelize the platform to later adopters.
-
 Perhaps most important is to treat the platform as a customer-facing product and recognize that its success is
 directly dependent on the success of its users and products; and as such it's
 vital that platform teams partner with app teams and other users to prioritize,
@@ -277,6 +269,14 @@ resistance and resentment from their users and miss a lot of the promised value.
 To counter this, platform teams should include product managers from the start
 to share roadmaps, gather feedback and generally understand and represent the
 needs of platform users.
+
+When adopting platforms, choosing the right capabilities and experiences to
+enable first, can be crucial. Capabilities that are frequently required and
+undifferentiated, like pipelines, databases and observability, may be a good
+place to start. Platform teams may also choose to focus first on a limited number
+of engaged and skillful app teams. Detailed feedback from such teams improves the
+first platform experiences; and people from those teams help champion and
+evangelize the platform to later adopters.
 
 Additionally, Platform teams can easily be overloaded in their efforts to support many
 product teams simultaneously. Ways to reduce load on the platform team include
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -259,6 +259,14 @@ following which implementers should keep in mind.
 1. Platform teams must seek support of enterprise leadership and show impact on
    value streams
 
+When adopting platforms, choosing the right capabilities and experiences to
+enable first, can be crucial. Capabilities that are frequently required and
+undifferentiated, like pipelines, databases and observability, may be a good
+place to start. Platform teams may also choose to focus first on a limited number
+of engaged and skillful app teams. Detailed feedback from such teams improves the
+first platform experiences; and people from those teams help champion and
+evangelize the platform to later adopters.
+
 Perhaps most important is to recognize that the success of a platform is
 directly dependent on the success of its users and products; and as such it's
 vital that platform teams partner with app teams and other users to prioritize,
@@ -279,6 +287,7 @@ teams. Ways to reduce load on the platform team include the following:
 1. Ensure platform teams are staffed appropriately for their domain and number
    of customers
 
+<<<<<<< HEAD
 Another challenge when adopting platforms is choosing the right capabilities and
 experiences to enable first. Services that are frequently required and
 undifferentiated, like pipelines, databases and observability, may be a good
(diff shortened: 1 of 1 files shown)


roberthstrand:
Could you follow the instructions to fix the DCO issue, @lianmakesthings? Details and instructions can be found [here](https://github.com/cncf/tag-app-delivery/pull/285/checks?check_run_id=10728873138). Thank you! 😄

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -274,6 +263,14 @@ following which implementers should keep in mind.
 1. Platform teams must seek support of enterprise leadership and show impact on
    value streams
 
+When adopting platforms, choosing the right capabilities and experiences to
```
joshgav, line 266: Would this fit better _after_ the following paragraph? One reason is that in the current order of the preceding list of challenges this item - "choosing the right priorities" - is no. 2. If you agree and move it let's add back a transition word to the beginning of the paragraph like "Another challenge when adopting platforms is...."

Or @lianmakesthings do you feel this item should be listed first? In that case let's switch the order of the preceding list too. In my opinion the existing order makes sense - treat platforms as products and gather user feedback first as that should influence priorities too.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -285,13 +282,15 @@ To counter this, platform teams should include product managers from the start
...
 needs of platform users.
 
-Another challenge when adopting platforms is choosing the right capabilities and
-experiences to enable first. Capabilities that are frequently required and
-undifferentiated, like pipelines, databases and observability, may be a good
-place to start. Platform teams may also choose to focus first on a limited
-number of engaged and skillful app teams. Detailed feedback from such teams
-improves the first platform experiences; and people from those teams help
-champion and evangelize the platform to later adopters.
+However, Platform teams can easily be overloaded in their efforts to support many
```
joshgav, line 285: Rather than "however", could we use a transition word which _follows on_ rather than _contradicts_ the previous? That is, "however" seems to say this is an exception to the previous statements, but I think it's in fact adding to them.

Could be something like "Another challenge to keep in mind is ensuring your platform team itself isn't cognitively overloaded! Ways to reduce load on the platform team include..."

joshgav (changes requested):
Thank you for this @lianmakesthings. Agree with moving the challenges for platform teams to the challenges section, have a couple requests re content itself, PTAL!

platforms-whitepaper/v1alpha1/paper.md
lianmakesthings: Going from my gut, I would probably keep this order and then change the list

lianmakesthings:
> Could you follow the instructions to fix the DCO issue, @lianmakesthings? Details and instructions can be found [here](https://github.com/cncf/tag-app-delivery/pull/285/checks?check_run_id=10728873138). Thank you! 😄

Will do.
But probably after we've agreed on the change requests first? They'll be closed if I rebase, I believe.

platforms-whitepaper/v1alpha1/paper.md
joshgav: That works, thanks.

platforms-whitepaper/v1alpha1/paper.md
joshgav: @lianmakesthings can you describe why you prefer the new order? Thanks!

I lean to the existing order because the top problem I've seen amongst early platform builders is that they choose and build capabilities without enough input from and iteration with product and application teams.  I've been surprised that these teams haven't successfully inculcated a product mindset despite all the literature on "platform as product".

I'd lean to putting "platform as product" first then following it with "prioritize carefully", cause only once a team has determined the right potential work items - i.e., by understanding customer needs - can they prioritize them.

joshgav:
@lianmakesthings 

> They'll be closed if I rebase, I believe.

I _think_ you can force-push and it'll stay open and just notify us. Something like `git commit -s --amend --no-edit && git push --force` - the `-s` signs the commit.

platforms-whitepaper/v1alpha1/paper.md
lianmakesthings: Sure! This is a very good discussion :)

The reason why I wanted to put this section about capabilities before the others was because in my mind the other two parts were diving deeper into the platform team and how they should be set up. 

Now that you defined the (now) second section as "platform as a product", I'd be inclined to agree with you on the order. 
But I would tease that out more in the text itself. To me, it only talks only about how the platform team should work with stakeholders. The idea of "productisation" is imho not obvious enough to people who are not familiar with the topic.

platforms-whitepaper/v1alpha1/paper.md
lianmakesthings, line 285:
```suggestion
Additionally, Platform teams can easily be overloaded in their efforts to support many
```

lianmakesthings:
```suggestion
Additionally, Platform teams can easily be overloaded in their efforts to support many
```

platforms-whitepaper/v1alpha1/paper.md
joshgav: I hear you. WDYT of the following suggestion to clarify the "platform as product" paragraph?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -274,6 +263,14 @@ following which implementers should keep in mind.
...
 
+When adopting platforms, choosing the right capabilities and experiences to
+enable first, can be crucial. Capabilities that are frequently required and
+undifferentiated, like pipelines, databases and observability, may be a good
+place to start. Platform teams may also choose to focus first on a limited number
+of engaged and skillful app teams. Detailed feedback from such teams improves the
+first platform experiences; and people from those teams help champion and
+evangelize the platform to later adopters.
+
 Perhaps most important is to recognize that the success of a platform is
```
joshgav, line 274: WDYT of expanding the first sentence here along these lines:

"Perhaps most important is to _treat the platform as a customer-facing **product** and recognize that its success_ is directly dependent on..."

Does this better connect this paragraph to the concept of "platform as a product"?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -285,13 +282,15 @@ To counter this, platform teams should include product managers from the start
...
 needs of platform users.
 
-Another challenge when adopting platforms is choosing the right capabilities and
-experiences to enable first. Capabilities that are frequently required and
-undifferentiated, like pipelines, databases and observability, may be a good
-place to start. Platform teams may also choose to focus first on a limited
-number of engaged and skillful app teams. Detailed feedback from such teams
-improves the first platform experiences; and people from those teams help
-champion and evangelize the platform to later adopters.
+Additionally, Platform teams can easily be overloaded in their efforts to support many
```
abangser, line 285: I really like this being moved down below the start of challenges. I think it reads really nicely as an addition to the challenges.

One concern I have though is if it ties into the three challenges that these paragraphs are trying to elaborate on. IMO "Platform teams must carefully choose their priorities and initial partner application teams" speaks to prioritising the roadmap, not the implementation.

Therefore, I would propose we keep the `## Enabling platform teams` header and move this section to actually after the whole `## Challenges with platforms`?

platforms-whitepaper/v1alpha1/paper.md
lianmakesthings:
```suggestion
Perhaps most important is to treat the platform as a customer-facing product and recognize that its success  is
```

platforms-whitepaper/v1alpha1/paper.md
lianmakesthings: @abangser does the below suggestion reflect what you mean?

```suggestion
## Enabling platform teams

Additionally, Platform teams can easily be overloaded in their efforts to support many
```

abangser: That is definitely half of it. I was just wondering if that change plus the paragraph would be better suited under the `Finally, it's vital...` paragraph. As in, does this section fit better as a wrap up or in the middle 🤔

lianmakesthings: Ah, I gotcha. Somehow I thought we were already at the end 😅 .

I agree, it would function nicely as a wrap-up. Which makes sense, since it was originally wrapping up the above section.
I'll send in another commit with some changed wording as new suggestion

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -301,6 +291,18 @@ demonstrate their direct impact on and relationships with product and value
 stream teams (see the previous two paragraphs), presenting the platform team as
 a strategic partner of product teams in delivering value to customers.
 
+## Enabling platform teams
```
lianmakesthings, line 294: @abangser what do you think of this?
I'm not quite happy with the wrap up here. The challenges section talks about productisation, capabilities, and organisational support and also offers some solutions.
Now we get to a section named "enabling platform teams" and it only talks about reducing load, but doesn't refer to any prior established point. It feels like something's missing.

Does that make sense?

abangser: That does make sense. What I hear (and agree with) is that this section feels like a return to the beginning of challenges rather than a nice wrap up that leaves the reader understanding challenges but also given opportunities to succeed. I think this is a byproduct of it previously being above the challenges section.

I started a suggestion below, but decided that keeping the conversation in this thread would be easier to follow. So please see this as a suggestion as a replacement for the paragraph in this section which tries to introduce the idea that there are support systems to the challenges that we have presented.

Please let me know if this helps with the concerns you have 🤞 

> It is clear from these challenges that platform teams are faced with a number of 
> diverse challenges that lead to cognitive load. Just as with their application team
> counterparts, this challenge grows with the number and diversity of users and 
> teams they need to support.
> 
> It is important to focus the platform team's energy on the experience and 
> capabilities that are unique to their specific business. Ways to reduce load on 
> the platform team include the following:
> 
> 1. Use implementations from managed service providers where reasonable
> 1. Leverage open source frameworks and toolkits for creating docs, templates and
>    compositions for application team use
> 1. Ensure platform teams are staffed appropriately for their domain and number
>    of customers

lianmakesthings: Yes, this reads much better in the context of the above section. I like it! :)

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -292,13 +292,18 @@ stream teams (see the previous two paragraphs), presenting the platform team as
 a strategic partner of product teams in delivering value to customers.
 
 ## Enabling platform teams
-Another challenge for platform teams is the amount of cognitive load, when attempting
-to support multiple teams simultaneously.
 
-Ways to reduce load on the platform team include the following:
+It is clear from these challenges that platform teams are faced with a number of
```
lianmakesthings, line 296: @abangser would you mind taking a look at this? It's pretty much what you drafted above, but I added some minor changes in wording.

abangser: That sounds great to me! Thanks for pushing for this and refining it Lian 🙇

joshgav:
@lianmakesthings this is great, thanks for your edits. Could you rebase on #289? I hope it doesn't effect you much. Then are you ready for this to be merged?

lianmakesthings:
> @lianmakesthings this is great, thanks for your edits. Could you rebase on #289? I hope it doesn't effect you much. Then are you ready for this to be merged?

Done and done.
I believe we've gone through all the suggestions & comments, so I'd say it's ready to merge.

joshgav:
Thank you @lianmakesthings!

Issue #278, opened 2023-01-10
[WG Platforms] Platforms Def - Measuring success of platforms


joshgav:
Review and finalize "How to measure success of platforms" section of Platforms whitepaper: <https://github.com/cncf/tag-app-delivery/blob/platforms-v1alpha1/platforms-whitepaper/v1alpha1/paper.md#how-to-measure-the-success-of-platforms>

Saim-Safdar:
Happy to work and lead this

taras:
I'm happy to support this

joshgav:
Thank you @Saim-Safdar and @taras, assigned.

salaboy:
Hi team, (@joshgav @jkleinlercher), this is something that I am covering in my book: Continuous Delivery for Kubernetes (http://mng.bz/jjKP) -> Chapter 9, and I built an open source example (https://github.com/salaboy/from-monolith-to-k8s/tree/main/four-keys) on DORA metrics (lead time, deployment frequency, mttr, change failure rate) in collaboration with the https://cd.foundation and the https://cdevents.dev team. I think that on this topic we can definitely collaborate and include some examples of how this can be tackled at the platform level.

joshgav:
Want to note that this "metrics" section is the only one IMO in the current revision that's not "content complete" and needs more than a review and edit.

For this section we need to do the following:

1. Verify that we've gathered and explained the right metrics and measures. Modify list as appropriate.
2. Add prose to help the section flow with the rest of the document.

@salaboy's links look promising, thank you!

joshgav:
> Frequency of updated deployments

We mentioned in a call today that we should define what a deployment is - does it include bug fixes, security fixes, dependency updates? Or perhaps only a featureful proper release?

abangser:
> we should define what a deployment is - does it include bug fixes, security fixes, dependency updates? Or perhaps only a featureful proper release?

Ah love this! IMHO a change is a change (so I would agree with your list as a reasonable starting point to describe a change).

I think one of the biggest issues is a platform that separates "hot fix" changes and "normal" changes (or some other naming convention). Basically everyone figures out ways to do a hot fix instead of the normal process as it is faster. If the release process is too slow to support a hot fix, it may take more effort, but the real fix is to the release process, rather than to introduce a fast track where you just end up with a second problem in the long run.

Saim-Safdar:
Some key points to consider here, while we are exploring at measuring success.

**Actionable accountability**

[...]

PR #297, opened 2023-02-07
[WG Platforms] Rewrite section on metrics

joshgav:
This is a rewrite of the section on metrics and measures based in part on discussions in #278. It categorizes meters an organization might track into the following three areas then provides some typical examples for each.

1. User satisfaction and productivity
1. User autonomy and process automation
1. Product and feature delivery

Here's the rendered version for easier review: <https://github.com/joshgav/tag-app-delivery/blob/0771a691c4a23b6cc00c1558efb6df3c8d58132c/platforms-whitepaper/v1alpha1/paper.md#how-to-measure-the-success-of-platforms>

@Saim-Safdar WDYT?

### DIFF
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -306,52 +306,63 @@ the platform team include the following:
 
 ## How to measure the success of platforms
 
-Having defined what platforms are and the values they intend to provide, let's
-discuss how to verify if a platform is indeed providing those values. The
-success of a platform depends on the success of its users, so these metrics
-should reflect a platform's impact on product teams that use it. We can measure
-a platform's impact in the following areas:
+Enterprises will want to measure whether their platform initiatives are
+delivering their anticipated values. Also, throughout this paper we've
+emphasized the importance of treating internal platforms as products, and good
+product management depends on quantitative and qualitative measurement of a
+product's performance. To meet these requirements, internal platform teams
+should gather user feedback and measure user activities.
+
+As with other aspects of internal platforms, though, platform teams should use
+the smallest viable effort to gather the feedback they need. We'll suggest
+metrics here but simple surveys or question and answer sessions may be most
+valuable initially.
+
+Categories of metrics that will help enterprises and platform teams understand
+the impact of their platforms include the following:
 
 1. User satisfaction and productivity
-1. User autonomy and process automation
-1. Product and feature delivery
 
-As we've said, the first value sought in many platforms is to improve and
-accelerate the work of their users, so measuring user satisfaction and
-improvements in users' productivity helps verify a platform's impact. Some
-popular metrics that reflect this include:
+The first quality sought by many platforms is to improve user experience in order
+to increase productivity. Metrics that reflect user satisfaction and
+productivity include the following:
 
-- Time for a new user to submit their first code changes to their product
-- "Net Promoter Score" (NPS) or other surveys that measure user satisfaction with a product
 - Active users and resources: number of capabilities provisioned and used from the platform
+- "Net Promoter Score" (NPS) or other surveys that measure user satisfaction with a product
 - Metrics for developer productivity such as those discussed in the SPACE framework [[3]]
 
-[3]: https://queue.acm.org/detail.cfm?id=3454124
+1. Organizational efficiency
 
-Another value sought in many platforms is to enable developer autonomy and
-self-service, reducing the number of manual steps and amount of human
-intervention required to acquire resources and services. So measuring the amount
-of time required to complete a process also shows the impact of a platform; consider
-measures such as these:
+Another benefit sought from many platforms is to efficiently provide common
+needs to a large user base. This is often achieved by enabling user self-service
+and reducing manual steps and required human intervention while implementing
+policies to guarantee safety and compliance. To measure the efficiency of a
+platform in reducing common work, consider measures such as these:
 
-- Latency to request and receive a space in a service, such as a database or namespace
-- Latency to build and deploy a complete service
+- Latency from request to fulfillment of a service or capability, such as a database or test environment
+- Latency to build and deploy a brand new service into production
+- Time for a new user to submit their first code changes to their product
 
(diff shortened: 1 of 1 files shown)


Saim-Safdar (commented):
Those are much-needed metrics for consideration, but I believe we need to highlight Outputs vs outcomes vs impact, thinking here is some teams don't use a metrics-driven approach instead using tracking and reporting. This could be caused by a lack of product thinking in platform teams or technical complexity in extracting the data (or both).

Setting an outcome-based goal (i.e. increasing the number of weekly querying users) for a specific quarter and expecting the team to deliver the outputs.

Another way to look into it is, the business KPIs your company sets won’t work immediately for your platform team. You’ll therefore need to find your own platform team KPIs, are we planning to cover or are we providing some insights into how to internalize and externalize KPIs for the platform team as the business KPI's changing or are these two separate things?

joshgav:
Thanks @Saim-Safdar. I hear you on considering objectives or outcomes alongside metrics. Do you think the Attributes section describes these higher order objectives? Maybe we can make it clear that the measures in this section represent attainment of the objectives described in the Attributes and Why sections. WDYT?

Thanks for bringing up business metrics versus platform metrics. The text proposed here nods to that in saying things like “the success of the platform depends on the success of its users”. Is there another contrast between platform and product objectives and metrics we could add? Thanks!

joshgav:
Hi @Saim-Safdar - I took another pass, focusing particularly on aligning measures with objectives (the attributes described earlier in the paper). I also chatted with @abangser and got some valuable tips and feedback. PTAL!

Reading your earlier comment again, you may be saying we should recommend enterprises or platform teams set quantitative outcome targets like "NPS > 50" or "MAU > 2000" rather than just tracking the metrics. If so, we could merge this as is yet still add that idea later.

abangser (approved):
Love it @joshgav ! In particular love that you one again point towards thinnest viable while also providing next step options.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -306,32 +306,63 @@ the platform team include the following:
...
-- Time for a new contributor to submit their first (or 10th!) PR
-- Number of contributions from users, e.g. Backstage plugins, Crossplane
-  compositions
-- Number of projects which diverge from "golden paths" and provided capabilities
-- Number of tickets opened related to platform capabilities
-- Developer satisfaction in the enterprise
-    - Net Promoter Score (NPS)
-    - SPACE framework: https://queue.acm.org/detail.cfm?id=3454124
+Enterprises will want to measure whether their platform initiatives are
+delivering their anticipated values. Also, throughout this paper we've
```
abhinavm-rafay, line 310: this is a little vague @joshgav can we clarify what we mean here by anticipated values?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -306,32 +306,63 @@ the platform team include the following:
...
+emphasized the importance of treating internal platforms as products, and good
+product management depends on quantitative and qualitative measurement of a
+product's performance. To meet these requirements, internal platform teams
+should gather user feedback and measure user activities.
```
abhinavm-rafay, line 314: should continuously gather

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -306,32 +306,63 @@ the platform team include the following:
...
+
+As with other aspects of internal platforms, though, platform teams should use
+the smallest viable effort to gather the feedback they need. We'll suggest
+metrics here but simple surveys or question and answer sessions may be most
```
abhinavm-rafay, line 318: one aspect here to talk about is also observing how users use your platforms - ethnographic studies these kind of studies can prove to be super useful

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -306,32 +306,63 @@ the platform team include the following:
...
+Categories of metrics that will help enterprises and platform teams understand
+the impact of their platforms include the following:
+
+1. User satisfaction and productivity
+
+The first quality sought by many platforms is to improve user experience in order
+to increase productivity. Metrics that reflect user satisfaction and
+productivity include the following:
+
+- Active users and resources: number of capabilities provisioned and used from the platform
```
abhinavm-rafay, line 330: should consider metrics here like DAU and MAU or monthly adoption rate as cited in the Palo Alto networks example

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -306,32 +306,63 @@ the platform team include the following:
...
+
+1. Product and feature delivery
+
+The ultimate objective of internal platforms is to deliver business value to
+customers faster, so measuring impact on a business's own product and feature
+releases demonstrates that the objectives of the platform are being met. The
+DevOps Research and Assessment (DORA) institute at Google suggests [[4]]
+tracking the following metrics:
+
+- Deployment frequency
```
abhinavm-rafay, line 354: here if we want to be more specific we can give examples such as 

- decrease in time from commit to deploy
- decrease in MTTR 
- decrease in change fail rate 
- increase in deployments

platforms-whitepaper/v1alpha1/paper.md
joshgav: I hear you! Yes, I'll change to "whether their platform initiatives are delivering the values and attributes discussed above". I'm thinking the Attributes section has the Objectives that inform the Key Results in this section.

platforms-whitepaper/v1alpha1/paper.md
joshgav: Yes, I hear you. Changing to "We'll suggest metrics here but simple surveys and analysis of user behavior may be most valuable initially."

platforms-whitepaper/v1alpha1/paper.md
joshgav: Well you won't know if they increase or decrease till you measure them!

joshgav:
Thanks for the feedback everyone! Still open to further iteration!

PR #298, opened 2023-02-08
Update paper.md

muemich:
* Added a glossary item about Stream Aligned Teams
* Minor changes and additions

Signed-off-by: Michael Mueller <michaelmuellerrw@gmail.com>

### DIFF
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -88,14 +88,16 @@ Enterprises adopting platform engineering can expect to achieve the following:
 
 1. Accelerate product development and delivery by intentionally dedicating teams
    to platform services and thereby reducing cognitive load on product teams
-1. Improve reliability and resiliency of products relying on platform
-   capabilities by dedicating experts to configure and manage them
-1. Accelerate product development and delivery by reusing and sharing platform
+2. Improve reliability and resiliency of products relying on platform
+   capabilities by dedicating experts to develop, configure and manage them
+3. Accelerate product development and delivery by reusing and sharing platform
    tools and knowledge across many teams in an enterprise
-1. Reduce risk of security, regulatory and functional issues in products and
+4. Easier and shorter onboarding of people moving between teams, as the platform 
+   to run on is the same 
+5. Reduce risk of security, regulatory and functional issues in products and
    services by governing platform capabilities and the users, tools and processes
    surrounding them
-1. Enable cost-effective and productive use of services from public clouds
+6. Enable cost-effective and productive use of services from public clouds
    and other providers by enabling delegation of implementations to those providers
    while maintaining control over user experience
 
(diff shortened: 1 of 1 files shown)


abangser (commented):
I think the small tweaks help a lot and that the description of stream align teams is great. Thanks for taking the time to PR!

Two questions before merging:
1. Should / can we go back to the list numbers using markdown where all are "1" meaning they can be reordered easily?
2. While I agree with the glossary addition, ado we need it? I think we removed this term from the paper?
2b. If we keep the glossary term, should / can we provide other names commonly used in place of stream aligned teams? E.g. Add something like "some organisations may call these product or delivery teams" or will that complicate things?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -88,14 +88,16 @@ Enterprises adopting platform engineering can expect to achieve the following:
 
 1. Accelerate product development and delivery by intentionally dedicating teams
    to platform services and thereby reducing cognitive load on product teams
-1. Improve reliability and resiliency of products relying on platform
-   capabilities by dedicating experts to configure and manage them
-1. Accelerate product development and delivery by reusing and sharing platform
+2. Improve reliability and resiliency of products relying on platform
+   capabilities by dedicating experts to develop, configure and manage them
```
joshgav, line 92: We didn't include "develop" because in some platform teams they don't actually construct the services and capabilities, rather they manage those provided by a cloud provider.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -88,14 +88,16 @@ Enterprises adopting platform engineering can expect to achieve the following:
...
+3. Accelerate product development and delivery by reusing and sharing platform
    tools and knowledge across many teams in an enterprise
-1. Reduce risk of security, regulatory and functional issues in products and
+4. Easier and shorter onboarding of people moving between teams, as the platform
```
joshgav, line 95: We'd considered this benefit to be part of the previous and mention it that way in the ensuing paragraph. I hear you though that it might be worth mentioning here. If so, let's make the language parallel to the other ones, something like: "Accelerate onboarding of current developers to new projects and teams by sharing the same platforms between them."

Can you suggest this in #305? Thanks!

joshgav (commented):
Thanks for your suggestions @muemich. 

Re "stream-aligned teams", I understand the idea behind that to be that enterprises are built around a core set of "value streams", which are workflows by which the enterprise delivers values to an economy. A "stream-aligned team" contributes directly to an enterprise's main value streams, in contrast to a platform team which contributes only indirectly.

I think we should leave out "value streams" and "stream-aligned teams" in this paper though and stick with "product", "application", or "workload" teams instead. The ideas behind value streams are fitting, but hopefully we capture them already using different terms in the "Why Platforms" section.

Thanks again!

muemich:
> 

re 1, sure. Most likely my IDE f... this up
re 2, the term was used in the paper and haven't had a reference or glossary item. If you remove it entirely then the glossary item isn't required.

platforms-whitepaper/v1alpha1/paper.md
muemich: I would argue that even the tailoring and configuration of cloud services is some sort of development. It it for sure developing a developer platform and I doesn't require writing actual code.

joshgav:
Hi @muemich and thanks again for these ideas. In the place in the doc where we use the term "stream-aligned" (the Evan Bottcher quote) we added a parenthetical '[product]' note next to it since as you noted we don't otherwise define the term. Check that out and see if it helps, and thanks for calling it out!

Other parts of the paper have changed significantly these past few weeks, so if you could read it over again and see if you still have the other feedback that would be great; perhaps you could open another PR with those changes. I'll go ahead and close this one as complete. Thanks again!

muemich:
Will read again when I'm back next week at a computer

Issue #273, opened 2023-01-10
[WG Platforms] Platforms Def - What is a platform


joshgav:
Review and finalize "What is a platform" section of Platforms whitepaper: <https://github.com/cncf/tag-app-delivery/blob/platforms-v1alpha1/platforms-whitepaper/v1alpha1/paper.md#what-is-a-platform>

abhinavm-rafay:
@joshgav it'd be great to potentially own this section if possible would love to add some details here as well as helping solicit feedback let me know if that works!

joshgav:
Thank you @abhinavm-rafay, that would be great, I went ahead and assigned you here; we can add others if appropriate.

As we chatted, let's try making significant edits via PRs to the `platforms-v1alpha1` branch, and having discussions here or in distinct issues as you feel appropriate. Thank you!

jkleinlercher:
> It seems like we have some doubts about the naming. We opted for talking about "platforms" in general, but as also understandable from Matthew Skelton's [comment](https://cloud-native.slack.com/archives/C020RHD43BP/p1673522113762099?thread_ts=1673388246.346779&cid=C020RHD43BP), that term is too generic and we probably need to be more specific.


> Therefore, an "application platform" (or "internal developer platform/IDP") would be a type of cloud native platform that supports the path to deliver value to the application end-users, from idea to production. But there are more.
> 
> Other types of cloud native platforms would be:
> 
> * core platform (containing only the services necessary to offer a Kubernetes-based environment with security, compliance and access control that could provide the foundation to build special-purpose platforms on top of it).
> * deployment platform (focused on the final part of the path to production)
> * data platform (oriented to big data or ML use cases)
> * observability platform (offering services to monitor, visualize, and work with telemetry and other signals coming from different sources, including other platforms).

I wonder if an "application platform" (which supports the path to deliver value to the application end-users, from idea to production) consists of a deployment platform and observability platform? The path from idea to production always needs deployment and observability, isn't it?

> These categories are not standard terminology, so I wouldn't try to use them here. At least, not in this first paper. I would talk more in general about "cloud native platforms" with focus on how we build them, both from a technical and an organisational point of view, and how to provide the best user experience. I'm not a fan of the "developer platform" terminology because application developers are not the only users of the platform. I think that the UX is critical for all users of the platform.

[...]

PR #301, opened 2023-02-09
What Is a Platform Modifications

abhinavm-rafay:
In regards to issue #273 address what is a platform including
- differentiation between IDP/service catalog and explaining that a platform doesn;'t mean just that or only that
- easier to understand examples
- evolution of platform in terms of product based on business needs/TVP
- explaining different stakeholders involved, not just devs

### DIFF
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -81,17 +81,19 @@ identity access, infrastructure operations, and app lifecycle.
 
 ## What is a platform
 
-A platform is a layer that provides common supporting capabilities and services 
-for many applications and use cases. Such a platform provides consistent user 
-experiences for getting, using and managing its capabilities and services, 
-including Web portals and pages, scenario-specific code templates, automatable 
-APIs and command-line tools.
+A platform for cloud-native computing is an integrated collection of
+capabilities defined and presented according to the needs of the platform's
+users. It is a cross-cutting layer that ensures a consistent experience for
+acquiring and integrating typical capabilities and services for a broad set of
+applications and use cases. A good platform provides consistent user experiences
+for using and managing its capabilities and services, such as Web portals,
+project templates, and self-service APIs.
 
 According to Atlassian [[1]], "platform teams create capabilities that can be 
-used by numerous stream-aligned teams with little overhead.... platform teams 
-minimize resources and cognitive load of the stream-aligned team... platform 
-teams can create a cohesive experience that spans across different user 
-experiences or products."
+used by numerous stream-aligned [product] teams with little overhead....
+platform teams minimize resources and cognitive load of the stream-aligned
+[product] team... platform teams can create a cohesive experience that spans
+across different user experiences or products."
 
 According to Martin Fowler and Evan Bottcher [[2]], "a digital platform is a 
 foundation of self-service APIs, tools, services, knowledge and support which 
@@ -99,58 +101,58 @@ are arranged as a compelling internal product. Autonomous delivery teams can
 make use of the platform to deliver product features at a higher pace, with 
 reduced coordination."
 
-Platforms are particularly relevant for cloud-native computing, which typically 
-separates supporting capabilities from application-specific logic more than 
-previous paradigms. In such environments resources like databases and object 
-stores, message queues and brokers, observability collectors and dashboards, 
-user directories and authentication systems, task runners and reconcilers and 
-more are managed independently and integrated into applications running in 
-containers and machines. A platform for cloud-native computing provides these to 
-many teams in ways that make them easy to integrate in applications and systems.
+The specific set of capabilities provided by a platform depends on the
+requirements of the platform's users. And while platforms _provide_ these
+required capabilities, it's critical to note that platform teams should not
+always _implement_ them themselves. Managed service providers or dedicated
+internal teams can maintain backing implementations while platforms are the
+thinnest reasonable layer that provides consistency for these implementations
+that meets an organization's requirements. For example, a very simple "platform"
+could be a wiki page with links to standard operating procedures to provision
+capabilities from providers, as described in [[3]].
 
-At its most basic, a platform provides consistent experiences to application 
-developers for acquiring and using individual services such as a database system 
-or a secret store. Example use cases an enterprise can meet with a basic platform 
-include the following:
+Because these platforms target no more and no less than an enterprise's internal
+users we often refer to them as _internal_ platforms.
 
-1. Developers of products and services can automatically provision runtime 
-   capabilities such as compute, storage, databases and service identity and 
-   immediately integrate and use those as part of their products
-1. Developers of products and services can automatically provision supporting 
-   services such as task runners, package registries, deployment orchestrators and 
-   observability systems to build, verify, operate and observe their products
-1. Operators of third-party products and services can automatically provision 
-   spaces and supporting services to deploy and use those third-party products and 
-   services
+Platforms are particularly relevant for cloud-native architectures because these
+separate supporting capabilities from application-specific logic more than
+previous paradigms. In cloud-like environments resources and capabilities are
+often managed independently and integrated with custom business components; such
+resources may include databases and object stores, message queues and brokers,
+observability collectors and dashboards, user directories and authentication
+systems, task runners and reconcilers and more. An internal platform provides
+these to enterprise teams in ways that make them easy to integrate in their
+applications and systems.
 
-A more advanced platform also _composes_ these capabilities into experiences and 
-templates fit for key scenarios like web application development or data analysis. 
-For example, an application developer platform can include templates for provisioning 
-complete development environments and could serve the following use cases:
+### Platform maturity
 
-1. Developers of products or services can automatically request a complete 
-   development environment to support iterative research and development of new 
-   features. This includes spaces in relevant services such as task runners and 
-   artifact storage, membership in designated teams, and publication of connection 
-   info such as URLs and secrets.
-1. Developers of products or services can use scenario-specific code and 
-   configuration templates to rapidly bootstrap, develop and deliver new products 
-   and features
-1. Stakeholders in a product or service can observe system and user behavior in 
-   those products and services via instrumentation, dashboards and alerts
+At their most basic, internal platforms provide consistent experiences for
+acquiring and using individual services such as a pipeline runner, a database
+system or a secret store. As they mature internal platforms also offer
+_compositions_ of such services as self-serviceable templates for key scenarios
+like web application development or data analysis, aka MLOps.
 
-A platform is bespoke to an organisation, supporting a unique set of users and 
-business needs. While application developement and delivery is often the first 
-capability a platform supports, it is important to focus on the individual needs of 
-the organization which often surfaces additional opportunities to optimise through 
-a platform offering. Some examples can include data operations, user management, and 
-incident management.
+Use cases an enterprise could meet with platforms might progress through the
+following:
 
-By offering consistent experiences for individual and/or scenario-oriented sets 
-of capabilities, platforms make it easy for their users to deliver valuable products.
+1. Product developers can provision capabilities on demand and immediately use
+   them to run systems, such as compute, storage, databases or identities.
+1. Product developers can provision service spaces on demand and use them to run
+   pipelines and tasks, to store artifacts and configuration, and/or to collect
+   telemetry.
+1. Third-party software operators can provision dependencies on demand and run
+   that software.
+1. Product developers can provision complete environments from templates
+   combining run-time and development-time services required for specific
+   scenarios, such as web development or MLOps.
+
+By offering consistent, compliant experiences for individual capabilities or
+sets of them, internal platforms ultimately make it easier and more efficient
+for their users to deliver valuable products.
 
 [1]: https://www.atlassian.com/devops/frameworks/team-topologies
 [2]: https://martinfowler.com/articles/talk-about-platforms.html
+[3]: https://teamtopologies.com/key-concepts-content/what-is-a-thinnest-viable-platform-tvp
 
 ## Attributes of platforms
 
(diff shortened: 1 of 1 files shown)


platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
-   observability systems to build, verify, operate and observe their products
-1. Operators of third-party products and services can automatically provision 
-   spaces and supporting services to deploy and use those third-party products and 
-   services
-
-A more advanced platform also _composes_ these capabilities into experiences and 
-templates fit for key scenarios like web application development or data analysis. 
-For example, an application developer platform can include templates for provisioning 
-complete development environments and could serve the following use cases:
+At its more basic, a platform is an **integrated product** developed by a
```
abangser, line 84: I really like these first two paragraphs as a way to introduce and expand on what internal means 🙌

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
+instead of focusing on software and app delivery.
+
+Therefore, the goal of the platform is to truly drive
+organizational efficiency and productivity by empowering internal stakeholders to 
+focus on their key responsibilities while giving them the tools and self-service 
+necessary to drive their day to day work. For developers for example, this could mean
+having the ability to deploy self-service environments for application development without 
+the overhead of understanding infrastructure components. Or for FinOps, this could
+mean being able to define chargeback rules without understanding getting too in depth of 
+how the metrics are actually collectedfrom cloud-native infrastructure.
```
abangser, line 111: I found this paragraph a bit confusing and difficult to read in the previous version as well. My concerns are:
1. The number of examples is difficult to parse
2. The use of cloud native here doesn't seem to lead to the challenges proposed.

Here is an attempt to capture what I mean:

```suggestion
It is said that today, every company is a technology company. In addition, the expectations on the availability and scalability of that software are growing. This has led to an increase in pressure on teams to manage more complex infrastructure like queues, high availability data stores, edge caches and more. For additional resiliency and scalability, this infrastructure is mainly built and managed on different physical machines from the software that depends on it. In order to configure and manage these resources developers are having to spend more time learning about infrastructure instead of focusing on software and app delivery.

A platform focuses on centralising the configuration and management of this infrastructure in order to enable organizational efficiency and productivity by empowering internal stakeholders to focus on their key responsibilities while giving application teams the tools and self-service. A team of platform experts not only reduces common work demanded of product teams but also optimizes platform capabilities used in those 
products. A platform team also maintains a set of conventional patterns, knowledge and tools used broadly across the enterprise ; enabling developers to quickly contribute to other teams and products built on the same 
foundations. The shared platform patterns also allow embedding governance and controls in templates, patterns and capabilities. Finally, because platform teams corral providers and provide consistent experiences bespoke to their unique needs over their offerings, they enable efficient use of public clouds and service providers for foundational but undifferentiated capabilities such as databases, identity access, infrastructure operations, and app lifecycle.
```

You will notice I pulled in the last paragraph of the why section as I think that we may want that here instead of in the above section?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
+It is key to note that a platform is bespoke to an organization, supporting 
+a unique set of users and business needs. Therefore, platforms are not to be confused 
+as just an Internal Developer Portal (IDP) or Service Catalog [[5]]. While those interfaces 
+can certainly be helpful for many organizations, ultimately platform engineering teams
+need to take a platform as product approach and think deeply about the business objectives 
+of their organizations. At its very basic level, a platform could simply be a place where 
+users go for basic actions like user management, technical docs, or incident management if that 
+is what is most needed.
+ 
+However, just like any product, a platform can evolve over time to meet more complex needs of an organization.
```
abangser, line 125:
```suggestion
However, just like any product, a platform must evolve over time to meet evolving needs of an organization. 
```

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
-1. Stakeholders in a product or service can observe system and user behavior in 
-   those products and services via instrumentation, dashboards and alerts
-
-A platform is bespoke to an organisation, supporting a unique set of users and 
-business needs. While application developement and delivery is often the first 
-capability a platform supports, it is important to focus on the individual needs of 
-the organization which often surfaces additional opportunities to optimise through 
-a platform offering. Some examples can include data operations, user management, and 
-incident management.
+1. Engineering Teams and SRE Teams have a central location for observability where infrastructure
```
abangser, line 133: I'd rather not get into more specific team names (e.g. SRE here)

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
+	and services can be onboarded into a single observability platform. 
+	This allows for a consistent approach to observability while giving developers 
+	an easy interface to be able to do this for any app or cloud service.
+1. Operators of third-party products and services can automatically provision 
+   spaces and supporting services to deploy and use those third-party products and 
+   services
+   
+Therefore, It is recommended that platform engineering teams start with the thinnest viable platform (TVP) [[6]],
```
abangser, line 141: Nice add of pointer to TVP

abangser (commented):
Just added a few points here, will need to revisit and may be worth touching base on a call later.

platforms-whitepaper/v1alpha1/paper.md
joshgav: The implicit question this paragraph is trying to answer is _why_ have "platforms" suddenly become important? The answer posited is that cloud-native computing's distributed systems paradigms lead to a need for platforms of capabilities.

We could make this question and answer more clear, perhaps along the lines of "platforms have become critical for cloud-native computing **because** supporting capabilities are no longer in line with business logic..."

The other items you all mention may fit best in the Attributes section.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
+how the metrics are actually collectedfrom cloud-native infrastructure. 
+
+Therefore according to Martin Fowler[[2]], **"a digital platform is a foundation of self-service APIs, tools, services, 
+knowledge and support which are arranged as a compelling internal product."**
+
+It is key to note that a platform is bespoke to an organization, supporting 
+a unique set of users and business needs. Therefore, platforms are not to be confused 
+as just an Internal Developer Portal (IDP) or Service Catalog [[5]]. While those interfaces 
+can certainly be helpful for many organizations, ultimately platform engineering teams
+need to take a platform as product approach and think deeply about the business objectives
```
joshgav, line 120: It would be best for us to focus on defining CNCF's opinion on platforms rather than countering other opinions - there will be no end to that! Can we take out the references to IDPs and service catalog here and instead list them in the Capabilities section where they fit? That would be enough to imply that we don't believe they're themselves a platform.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
+	This allows for a consistent approach to observability while giving developers 
+	an easy interface to be able to do this for any app or cloud service.
+1. Operators of third-party products and services can automatically provision 
+   spaces and supporting services to deploy and use those third-party products and 
+   services
+   
+Therefore, It is recommended that platform engineering teams start with the thinnest viable platform (TVP) [[6]], 
+which is very similar to the concept of an MVP - solve
+for the key needs of your organization with the most basic and continue to evolve your platform just like 
+any other product.
```
joshgav, line 144: Could we integrate this into a single paragraph on TVP and "internal"?

platforms-whitepaper/v1alpha1/paper.md
joshgav, line 133: Let's agree on a set of use cases together and then write them up to revise this list.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
+which is very similar to the concept of an MVP - solve
+for the key needs of your organization with the most basic and continue to evolve your platform just like 
+any other product. 
 
 By offering consistent experiences for individual and/or scenario-oriented sets 
 of capabilities, platforms make it easy for their users to deliver valuable products.
 
 [1]: https://www.atlassian.com/devops/frameworks/team-topologies
 [2]: https://martinfowler.com/articles/talk-about-platforms.html
+[3]: https://thenewstack.io/platform-as-a-product-true-devops/
```
joshgav, line 151: I'm not yet sure how we'll do footnotes or if we want to highlight specific vendors and posts. In the meantime can we add these to the list at the end of the doc?

platforms-whitepaper/v1alpha1/paper.md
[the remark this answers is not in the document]
abhinavm-rafay: removed the word integrated here

platforms-whitepaper/v1alpha1/paper.md
abhinavm-rafay: fixed and included some notes about TVP and how that can evolve into an internal platform

platforms-whitepaper/v1alpha1/paper.md
abhinavm-rafay: addressed based on comments in last meeting - happy to talk through this more

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,77 @@ identity access, infrastructure operations, and app lifecycle.
...
-   services
-
-A more advanced platform also _composes_ these capabilities into experiences and 
-templates fit for key scenarios like web application development or data analysis. 
-For example, an application developer platform can include templates for provisioning 
-complete development environments and could serve the following use cases:
+At its more basic, a platform is an **integrated product** developed by a 
+platform engineering team that is custom built and composed of supporting 
+capabilities and services designed to reduce the cognitive load and burden 
+experienced by **internal** stakeholders of an enterprise organization
```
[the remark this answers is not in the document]
joshgav: While a platform may indeed often be a product, I don't think that's _intrinsic_ to its definition. That is, you _could_ have a platform that's not a product; in fact I'd argue that we emphasize the idea of "platform as product" because it's really _not_ a product - it's an internal utility but for it to succeed it must be treated _like_ a product.

I'd prefer we stick with the original 3 paragraphs for this section.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,135 +81,113 @@ identity access, infrastructure operations, and app lifecycle.
...
-complete development environments and could serve the following use cases:
+At its more basic, a platform is a **product** developed by a 
+platform engineering team that is custom built and composed of supporting 
+reusable capabilities and services designed to reduce the cognitive load and burden 
+experienced by **internal** stakeholders of an enterprise organization 
+while still safely shipping software at speed [[4]]. 
+
+Note that while internal stakeholders typically refer to developers, they can 
+refer to DevOps, SREs, FinOps, Customer Success, and anyone who is responsible 
+for software delivery and cloud native infrastructure in an enterprise organization [[2]].
```
joshgav, line 92: I think we can skip this side note here for the sake of being concise.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,135 +81,113 @@ identity access, infrastructure operations, and app lifecycle.
...
+
+The goal of any platform should be to drive 
+organizational efficiency and productivity by empowering internal stakeholders to 
+focus on their key responsibilities while giving them the tools and self-service 
+necessary to drive their day to day work. In order to discover those needs, a platform as product
```
joshgav, line 97: This sentence is a helpful transition, thanks! I like how you describe TVP as a response to meeting the most basic needs of users.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,135 +81,113 @@ identity access, infrastructure operations, and app lifecycle.
...
+approach is needed which means starting off with building only what is necessary and then evolving 
+the platform over time with user feedback and usage metrics. For example, a "platform" may start off as a wiki page 
+with links to provision each required capability. Some [[3]] have referred to this principle as seeking
+the "thinnest viable platform" that meets an enterprise's requirements. 
+
+As the user needs of internal stakeholders evolve in an organization and platform owners get more used feedback
+and success metrics, platform engineering teams in an enterprise must take ownership of delivering the required capabilities. 
+This typically leads into the evolution of the thinnest viable platform into an **internal platform** in cloud-native computing
```
joshgav, line 105: I don't think TVP _leads_ to internal platforms. Rather internal platforms start out as thinly as possible. TVP is really only applicable to an internal platform - an external platform by definition has to cater to many organizations' needs.

I've suggested TVP is related to "internal" but I'm still thinking :thinking: how to express that and work in "internal" here!

joshgav (commented):
Thanks Abhinav for your updates! I've made a few comments and put up #308 for you to consider too :laughing:.

joshgav:
I realize I need to be clearer on my objections to changing the first paragraph 😂. This PR proposes to add the following to the opening sentence which I don't think we should. We could make this paragraph clearer though.

Should we add something like:

"A platform is a collection of capabilities defined and presented based on the needs of its users."

Currently this PR adds:

- is an **integrated product**
- developed by a platform engineering team
- that is custom built

My main objection is that being a "product" isn't part of the definition of a platform (i.e.
, "a collection of capabilities..."), rather it's an attribute of a good one. The latter two statements aren't universally true.

platforms-whitepaper/v1alpha1/paper.md
abangser: Maybe this could be an entry/clarification in the glossary? We already have `Platform users`, so if we are missing a few examples we could add them there?

platforms-whitepaper/v1alpha1/paper.md
abangser: Just want to see if we are on the same page with TVP as not just how something starts, but how it continues. I believe the industry already has a term for starting out small and that would be MVP (Minimum Viable Product). My understanding is that TVP evolves that to say, in the world of platforms, it is best to stay thin and continue to add value over as many widely available tools as possible.

So to expand further (since i won't be around to clarify 😅:
A thin platform may start minimal (only covering a few use cases) but grow past minimal by covering many use cases while staying thin by continuously looking for ways to outsource capabilities that were originally innovative and custom but have since become commodity in the industry.

platforms-whitepaper/v1alpha1/paper.md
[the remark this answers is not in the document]
abangser: This is a tough one because I think that our group wants to advocate for the platform to be considered a platform while in reality that is still quite far to the right on the adoption curve.

By adding it here on the first line we are surely setting a stake in the ground which could either be very good way to build a shared assumption with the reader or could risk alienating. I am quite torn 🤔

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,135 +81,113 @@ identity access, infrastructure operations, and app lifecycle.
...
-1. Stakeholders in a product or service can observe system and user behavior in 
-   those products and services via instrumentation, dashboards and alerts
-
-A platform is bespoke to an organisation, supporting a unique set of users and 
-business needs. While application developement and delivery is often the first 
-capability a platform supports, it is important to focus on the individual needs of 
-the organization which often surfaces additional opportunities to optimise through 
-a platform offering. Some examples can include data operations, user management, and 
-incident management.
+1. Finops and other internal stakeholders get access to a dashboard for managing costs across business units with the ability to define chargeback rules without having to understand the underlying compute infrastructure
```
abangser, line 126: IMO this behaviour is FinOps, not the person being FinOps. I also fear that FinOps is a bit of a buzz word. We are already buzzwording a lot in our core domain (and trying to explain those words as we go!), so I would stay away from buzzwords that do not explicitly move forward platform engineering.

Also, my read is that stakeholders in this case are budget holders which naturally all roll into the CFO office (which is that FinOps?).

```suggestion
1. Budget holders get access to a dashboard for managing costs across business units with the ability to define chargeback rules without having to understand the underlying compute infrastructure 
```

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,135 +81,113 @@ identity access, infrastructure operations, and app lifecycle.
...
-capability a platform supports, it is important to focus on the individual needs of 
-the organization which often surfaces additional opportunities to optimise through 
-a platform offering. Some examples can include data operations, user management, and 
-incident management.
+1. Finops and other internal stakeholders get access to a dashboard for managing costs across business units with the ability to define chargeback rules without having to understand the underlying compute infrastructure 
+1. Operators of third-party products and services can automatically provision 
+   spaces and supporting services to deploy and use those third-party products and 
+   services
+
+These examples illuminate the definition of a platform according to Martin Fowler[[2]],
```
abangser, line 131: This was a quote from Bottcher on Fowler's blog. I have confirmed we cite correctly in the original, but I would be uncomfortable with only citing Martin here.

In addition, I fear that without the lead in of the quote, this is a bit of a throw away rather than the punchy statement that it was originally provided as. Was there a specific goal in mind by removing the industry quotes from the start of this section (both Atlassian and Bottcher/Fowler)?

abangser (commented):
I am not going to be super available for the next few weeks so please don't block on a reply from me! I think that this is a really interesting set of discussions and I look forward to seeing how this and #308 work together to refine the section!

platforms-whitepaper/v1alpha1/paper.md
ThomasVitale: I think the glossary would be a good place for that. Either way, I would consider listing user roles like testers, operators, and managers as examples of users. Otherwise we would probably need to explain what is FinOps, what is Customer Success, or what is DevOps (this one in particular could explode very quickly 😊).

platforms-whitepaper/v1alpha1/paper.md
ThomasVitale: I had the same concern about the quote. The author of that article is Evan Bottcher, published on Martin Fowler's blog.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -81,76 +81,64 @@ identity access, infrastructure operations, and app lifecycle.
...
+the platform over time with user feedback and usage metrics. For example, a "platform" may start off as a wiki page 
+with links to provision each required capability. Some [[3]] have referred to this principle as seeking
+the "thinnest viable platform" that meets an enterprise's requirements. 
+
+As the user needs of internal stakeholders evolve in an organization and platform owners get more used feedback
+and success metrics, platform engineering teams in an enterprise must take ownership of delivering the required capabilities. 
+This typically leads into the evolution of the thinnest viable platform into an **internal platform** in cloud-native computing 
+paradigms. 
+
+Internal platforms are particularly relevant for cloud-native architectures because these
```
ThomasVitale, line 108: Small suggestion:

```suggestion
Internal platforms are particularly relevant for cloud-native architectures because they
```

PR #304, opened 2023-02-10
[WG Platforms] Remove irrelevant sections from charter

joshgav:
This removes sections from the WG Platforms charter which were only relevant at the time of its creation, such as founding members, current chairs and sponsors, and meeting times. Items which are still relevant now are on the [WG home page](https://github.com/cncf/tag-app-delivery/tree/main/platforms-wg).

Fixes #303 and fixes #296.


roberthstrand (approved):
👍

PR #305, opened 2023-02-14
Platform attributes: optimize wording and format

ThomasVitale:
- Format the attributes to combine headline and explanation for better readability and clearer understanding.
- Combine self-service and automation since they share the explanation and foundation.
- Add "secure by default" which was discussed in the gdoc but not ported over to GitHub

### DIFF
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -183,7 +183,7 @@ of a platform.
    as a golden path.
 1. **Self-service**. A platform should be self-serviceable. Users must be able to request and
    receive capabilities autonomously and automatically. This property is key to allowing a platform
-   team to enable multiple stream-aligned teams at scale. The platform capabilities should be
+   team to enable multiple stream-aligned teams and scale as needed. The platform capabilities should be
    available on demand and with minimal manual intervention via the interfaces described above.
    For example, it should be possible for a user to request a database and receive its locator
    and credentials by running a command-line tool or filling out a form on a web portal.
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -154,62 +154,55 @@ of capabilities, platforms make it easy for their users to deliver valuable prod
 
 ## Attributes of platforms
 
-Following are key attributes that affect the success of a platform for
-cloud-native computing. Such a platform should:
+After defining what a platform is and why an organization might want to build one,
+it's relevant to identify what are the key attributes that affect the success
+of a platform.
 
-1. Be designed and continuously improved based on requirements of its users like
-   any product
-1. Manage common capabilities required and used by many application teams
-1. Offer consistent interfaces and experiences - GUIs, APIs, and tools - for
-   requesting, managing and observing capabilities
-1. Include documentation and code templates to ease use of its capabilities.
-   Advanced platforms should offer docs and templates for composed scenarios.
-1. Be self-serviceable - users must be able to request and receive capabilities
-   and compositions independently
-1. Be automatable - users must be able to program requests for capabilities
-1. Encapsulate implementation of capabilities and services while enabling
-   required configuration and observation
-1. Be optional - users should be able to use some platform capabilities but
-   acquire others outside of the platform's offerings
-
-A platform exists to serve the common requirements of many disparate users \[1,2\]
-and use cases. It serves those requirements through consistent interfaces of
-several types \[3\], particularly HTTP APIs, Web portals, and command-line tools.
-For example, all platform users require the ability to observe their workloads,
-so a platform might offer a capability to collect and display logs, metrics and
-traces from applications. It might enable users to request and use that
-capability via an API and a Web page.
-
-To be able to use a platform's offerings, users require documentation and
-examples \[4\]. So a platform team needs to not only make a capability like
-observability available, but also publish templates and docs to help users
-consume it. For example, a platform team might publish a reusable workflow for
-building, testing, deploying and verifying a Web application on Kubernetes.
-Beyond examples for individual resource types, advanced platforms should group
-collections of templates and guidance for specific scenarios, a bundle often
-described as a golden path.
-
-The capabilities provided by a platform should be available to users on demand
-\[5\] with minimal manual intervention. The interfaces discussed above - APIs and
-web pages - should enable users to get resources they need in a few minutes with
-at most an approval. For example, a web page describing a database management
-service should include a button to request a database and share a locator and
-credentials for that database within a few minutes of the request.
-
-One of the purposes of a platform is to reduce cognitive load on product teams,
-so services offered by the platform should not need to be operated by consumers;
-either the platform provider or its delegates, such as a cloud service, should
-handle those operations, hiding complexity \[7\] from users. For example, users
-may often require a message queue, but shouldn't have to manage a message
-broker.
-
-Platforms are intended to make product development more efficient, so it's
-important they not be an impediment \[8\] when the most efficient path is to not
-use the platform's offerings. Platforms should enable product teams to
-provide and manage their own capabilities when necessary. For example, if a
-platform team doesn't provide a graph database and it's required for a product,
-it should be possible for that product team to provision and operate a graph
-database themselves.
+1. **Platform as a product**. A platform exists to serve the requirements of its users
+   and it should be designed and evolved based on those requirements, similar to any
+   other software products.
+1. **Common requirements**. A platform should provide the necessary capabilities to
+   support the most common use cases across stream-aligned teams, and prioritize those
+   over more specific capabilities that are only used by a single team to maximize
+   the value delivered.
+1. **User experience**. A platform should offer its capabilities through consistent
+   interfaces and focus on the user experience. Users should be able to request, manage,
+   and observe such capabilities using a combination of GUIs, APIs, command-line tools,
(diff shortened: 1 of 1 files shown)


ThomasVitale:
@joshgav I reorganised the content of the section to combine headline and explanation for a better flow and clarify. I have also added the "secure by default" point we discussed a while ago on GDocs. What do you think?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -154,62 +154,55 @@ of capabilities, platforms make it easy for their users to deliver valuable prod
...
-database themselves.
+After defining what a platform is and why an organization might want to build one,
+it's relevant to identify what are the key attributes that affect the success
+of a platform.
+
+1. **Platform as a product**. A platform exists to serve the requirements of its users
+   and it should be designed and evolved based on those requirements, similar to any
+   other software products.
+1. **Common requirements**. A platform should provide the necessary capabilities to
+   support the most common use cases across stream-aligned teams, and prioritize those
```
abangser, line 165: > stream-aligned teams

Do we define this anywhere? I love Team Topologies, but this is not a globally known concept (yet!).

*Note: Of course this is a question for the paper and not this specific PR, just want to call attention as I spotted it here!

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -154,62 +154,55 @@ of capabilities, platforms make it easy for their users to deliver valuable prod
...
+   over more specific capabilities that are only used by a single team to maximize
+   the value delivered.
+1. **User experience**. A platform should offer its capabilities through consistent
+   interfaces and focus on the user experience. Users should be able to request, manage,
+   and observe such capabilities using a combination of GUIs, APIs, command-line tools,
+   IDEs integrations, and portals, depending on their needs. For example, a platform
```
abangser, line 171: > Users should be able to request, manage,
> and observe such capabilities using a combination of GUIs, APIs, command-line tools,
> IDEs integrations, and portals, depending on their needs

This seems to challenge our points about thinnest viable platform, though I think the intention is spot on. I wonder we can shorten to just the first sentence? Or if we can some how couch the remaining explanation as a bigger endeavour than some platforms need?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -154,62 +154,55 @@ of capabilities, platforms make it easy for their users to deliver valuable prod
...
+   examples. A platform should be delivered with proper documentation addressing the
+   needs of its users. It should also provide tools to accelerate the onboarding of new projects
+   that can help users consume the necessary platform services in a quick and simple way.
+   For example, the platform could offer a reusable supply chain workflow for building, scanning,
+   testing, deploying, and observing a web application on Kubernetes. Such a workflow could be
+   offered with an initial project template and documentation, a bundle often described
+   as a golden path.
+1. **Self-service**. A platform should be self-serviceable. Users must be able to request and
+   receive capabilities autonomously and automatically. This property is key to allowing a platform
+   team to enable multiple stream-aligned teams at scale. The platform capabilities should be
```
abangser, line 186: This sentence sounds a bit repetitive as is. With this edit I am trying to highlight the difference of handling current needs and scaling in the future.

```suggestion
   team to enable multiple stream-aligned teams and scale as needed. The platform capabilities should be
```

abangser (commented):
I think this is a really nice change.

One byproduct of this formatting is that it comes across as more a definitive list with specific, definable, labels. The clarity and succinctness that the bolded list headers brings, also breeds an expectation of accuracy and completeness.

I will think on this a bit myself as I am not sure which intent I would have expected or preferred!

platforms-whitepaper/v1alpha1/paper.md
roberthstrand: Do we feel that the wording is so self-explanatory that we could use it even if the reader doesn't know Team Topologies?

platforms-whitepaper/v1alpha1/paper.md
roberthstrand: I agree. I think the first sentence is enough to get the point across, while listing everything one could build for a better DX can quickly become bloated.

roberthstrand:
@ThomasVitale, could you run through your commit and sign-off, so that the DCO check works out? Details: https://github.com/cncf/tag-app-delivery/pull/305/checks?check_run_id=11384801820

joshgav (changes requested):
This is a great update, thank you @ThomasVitale. I like how you merged the contents as headings, it feels like the reader will follow this better.

I lean to agree with @abangser that we shouldn't introduce the term "[value] stream" at this point; let's see if we can use "product" or "application" instead, to reduce cognitive load on our readers! :laughing:

I'm also going to think about the phrases in the headings like "Cognitive load" and "Common requirements", if there might be a way to make them more expressive. :thinking: 

Thanks!

platforms-whitepaper/v1alpha1/paper.md
ThomasVitale: Good point! I wasn't sure about this, and then I saw we were adding the definition in https://github.com/cncf/tag-app-delivery/pull/298/files and went with "stream-aligned teams". But I agree that "product teams" would be better in this context since it's more self-explanatory. I'll change that.

platforms-whitepaper/v1alpha1/paper.md
ThomasVitale: Should I keep the first sentence, remove the second one, and then leave the example? Or would you prefer leaving the example off as well?

abangser: I think the last sentence is great where it brings to life the concept of meeting platform users where they are. Maybe a trimmed down version would be:

> 1. **User experience**. A platform should offer its capabilities through consistent
   interfaces and focus on the user experience. Platforms should endeavor to meet their users where they are which may mean different interfaces such as IDEs; command-line tools, chat bots, or web portals.

joshgav: +1 to the change and something like Abby's refactor, but Abby I'm not sure if you're suggesting removing the following sentence with an example or not? I'm okay with keeping it, it gives a bit of color; but okay with removing it too if we feel it's repetitive.

ThomasVitale:
@roberthstrand I should have fixed the signoff now

platforms-whitepaper/v1alpha1/paper.md
ThomasVitale: I have replaced "stream-aligned teams" with "product teams".

platforms-whitepaper/v1alpha1/paper.md
ThomasVitale: I have rephrased the paragraph based on your suggestions. What do you think?

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -154,62 +154,55 @@ of capabilities, platforms make it easy for their users to deliver valuable prod
...
-Platforms are intended to make product development more efficient, so it's
-important they not be an impediment \[8\] when the most efficient path is to not
-use the platform's offerings. Platforms should enable product teams to
-provide and manage their own capabilities when necessary. For example, if a
-platform team doesn't provide a graph database and it's required for a product,
-it should be possible for that product team to provision and operate a graph
-database themselves.
+After defining what a platform is and why an organization might want to build one,
+it's relevant to identify what are the key attributes that affect the success
+of a platform.
```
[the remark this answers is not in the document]
ThomasVitale: I have updated that paragraph as suggested.

ThomasVitale:
@abangser @joshgav I thought a bit about the impression of the bullet list with headings in bold. I don't know how much it helps, but I replaced the phrasing "THE key attributes" with "SOME key attributes". Perhaps we can extend the initial paragraph to make it even clearer that these are just examples of attributes and not a definitive list?

Also, as @joshgav pointed out, the "Common requirements" and "Cognitive load" headings were not very expressive. I thought about it and want to propose removing the first one to make it part of the "Platform as a product" bullet, since that's basically part of it: good product management is also about focusing on the high-value requirements first that will benefit the most users. I also replaced "Cognitive load"  with "Reduced cognitive load for users". What do you think?

platforms-whitepaper/v1alpha1/paper.md
abangser: love it Thomas!

abangser:
I really like the updates you've made @ThomasVitale 👍

ThomasVitale:
Thanks a lot, @abangser!

joshgav:
Thanks @ThomasVitale! The adjustment from "the" to "some" makes sense, thanks. I'm a bit less sure about integrating "platform as product" with "focus on common requirements" but am willing to go with it.

Re mentioning other attributes/objectives/values expected from internal platforms, we do want this paper to evolve based on feedback. Maybe a good approach is to couch the whole paper when released in a call to action to give us feedback and help us iterate, let's discuss in our next call.

joshgav (approved):

muemich (changes requested):
Add "Accelerate onboarding of current developers to new projects and teams by sharing the same platforms between them."
To: https://github.com/cncf/tag-app-delivery/pull/305/files#diff-defb6c1a2cbd28999ffdb84871bd5e6b34547d55f8e3bb88deebe835b829eee2L63

joshgav:
Thank you @ThomasVitale! @muemich can you rebase your proposal on this latest change? Thanks!

muemich:
Sure, will take a bit as I'm currently on the road.

Issue #280, opened 2023-01-10
[WG Platforms] Platforms Def - Glossary


joshgav:
Review and finalize "Glossary" section of Platforms whitepaper: <https://github.com/cncf/tag-app-delivery/blob/platforms-v1alpha1/platforms-whitepaper/v1alpha1/paper.md#glossary>

joshgav:
Thank you for #307, #298 and other discussions.

PR #307, opened 2023-02-17
First glossary review

ugho16:
Here's my first review of the glossary. Since it's my first contribution, I have kept it small: from your feedback, I'll be able to better understand the direction you want to give to this section.
Let me know if it can work, or how I can further extend it and keep contributing!

P.S. While checking out from the `platforms-v1alpha1` branch, git kept an edit I did not actually make on the 2021-04-15-chat.txt file. As you can see, I did not change anything, maybe it was some issue related to LF.


ugho16:
PR related to #280

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -465,21 +465,38 @@ influential.
 
 **Platforms** enable developers and operators to deliver applications and
 services faster by providing and managing common capabilities. Platforms bridge
-between platform users and platform capability providers.
+between platform users and platform capability providers, and are developed and
```
joshgav, line 468: Do you think "developed" implies writing code? Would we prefer not to imply that platforms require writing code?

joshgav (approved):
I like the change in order, the section flows better now, thanks @ugho16! The "platform teams" paragraph wraps that story up nicely too. Also like how you included TVP (check it out @abhinav-rafay)!

platforms-whitepaper/v1alpha1/paper.md
ugho16:
```suggestion
between platform users and platform capability providers, and are built and
```

Good point. We can use "built", if that sounds better for you.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -465,21 +465,38 @@ influential.
...
 services faster by providing and managing common capabilities. Platforms bridge
-between platform users and platform capability providers.
+between platform users and platform capability providers, and are built and
+maintained by platform teams.
+
+**Platform capability providers** develop and maintain the capabilities offered
+by the platform. Providers can be both external organizations or internal teams,
+and capabilities can be infrastructure, runtime, or other supporting services.
+
+**Platform developers** and **platform engineers** are responsible for developing and
```
ThomasVitale, line 475: Are these two terms used as synonyms or is there a difference we are trying to convey? Could we keep only one of the two?

ugho16: I found both terms in the paper, and I decided to keep them both here as well for the sake of consistency. The term "engineers" is more inclusive as it includes also people who don't write code, while "developers" is more code-related.

We can choose to adopt only one of them: in this case, I suggest reviewing the entire paper to apply the change consistently. I can take care of this if needed.

roberthstrand: Both work depending on the organizations structure, but I would agree that using engineer is a good default to use.

ThomasVitale: I would also lean towards "platform engineers" and use that consistently throughout the paper.

joshgav: +1 from me, might as well use "platform engineer" here in line with the industry buzz. I think it's okay to use "developer" as a synonym in other places in the doc though to keep it interesting.

platforms-whitepaper/v1alpha1/paper.md
```diff
@@ -465,21 +465,38 @@ influential.
...
-**Platform capability providers** implement infrastructure services, often as
-controllers and operators.
-
-**Platform developers** design interfaces and tools to enable provisioning and
-integration of platform capabilities in applications.
-
-**Platform product managers** are responsible for understanding the experience of
-platform users and building a roadmap that addresses platform product gaps and
-opportunities.
+**Thinnest viable platform (TVP)** is a concept originally defined in the book *Team Topologies*,
```
joshgav, line 496: nit - you can take out the comma after "Team Topologies"

joshgav:
@ugho16 can you make the final changes and then we'll merge? Thanks!

joshgav:
@ugho16 I'm going to merge this then make the suggested change by @ThomasVitale in #310.

ugho16:
@joshgav thank you so much, sorry I couldn't do it sooner

Issue #279, opened 2023-01-10
[Platforms] Platforms Def - Capabilities


joshgav:
Review and finalize "Capabilities of platforms" section of Platforms whitepaper: <https://github.com/cncf/tag-app-delivery/blob/platforms-v1alpha1/platforms-whitepaper/v1alpha1/paper.md#capabilities-of-platforms>

joshgav:
@taras gave feedback that we may use the term "capabilities" so much in the paper that it's distracting. I'll take a bit of responsibility for that, I often see the core of a platform as a collection of (self-serviceable) capabilities; I even explored how to build a platform from this perspective here: <https://blog.joshgav.com/posts/deliver-platform-capabilities>

Taras also shared that in some platforms the term "feature" may be used with the same intent as we're using "capabilities".

I'll do a review and see if we can reduce use of the term, find good synonyms and improve readability.

Comparing the terms "feature" and "capability" I favor "capability" cause it reflects the nature of the thing itself - something that adds a capability to an application. "Feature" to me means more a generic part of any product.

Thoughts? Thanks!

jkleinlercher:
For me a capability is something bigger than a feature. A capability consists of multiple features.

roberthstrand:
> For me a capability is something bigger than a feature. A capability consists of multiple features.

I agree. Capability might be one or more features combined, and a feature might not necessarily be thought of as a capability.

E.g. secrets management is a capability that could consist of features like encryption-at-rest, automatic access to secret for platform workloads, secret sharing with external users, and so on.

joshgav:
@roberthstrand  @jkleinlercher 

> For me a capability is something bigger than a feature. A capability consists of multiple features.

> E.g. secrets management is a capability that could consist of features like encryption-at-rest, automatic access to secret for platform workloads, secret sharing with external users, and so on.

I like this - just like a product has features, a platform capability has features. Perhaps we could define "features" as "**properties of the implementation of a capability**", WDYT? I'll try an edit pass to add this idea.

Can folks share links to other descriptions and opinions on terms like these? Thanks.

joshgav:
I'd like to volunteer for editing this section (and assigned myself here in GitHub). I'm passionate about describing and distinguishing the true nature of capabilities to help teams build a **thorough** yet **streamlined** platform layer: _thorough_ in that all capabilities apps and devs require are present, and _streamlined_ in that the focus is on providing a conceptual capability over deploying many implementations.

[...]

PR #310, opened 2023-02-28
Added sustainability to platform capabilities

ugho16:
Hello everyone, and apologies for the late contribution. I added sustainability and cost governance to the "Capabilities of platforms" paragraph. 
_Cost governance_ was mentioned earlier in the paper but did not appear in the capabilities section. 
_Sustainability_, instead, was not mentioned before, but it will be a very hot topic in the near future and it can also become a decision driver. The CNCF itself started a working group on sustainability, so I think it would be nice to refer to the topic.

If you think this can be a good point and if you know other tools that can fit in this category, please feel free to add them.

### DIFF
--- a/platforms-whitepaper/v1alpha1/paper.md
+++ b/platforms-whitepaper/v1alpha1/paper.md
@@ -409,6 +409,7 @@ platform is complete.
    policy enforcement
 1. **Artifact storage**: container image and language-specific package management,
    custom binaries and libraries, source code
+1. **Sustainability and cost governance** services and products
 
 <table>
   <thead>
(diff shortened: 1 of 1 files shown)


joshgav:
Thanks for raising these @ugho16.

To this point we've considered watching costs a form of monitoring/observability. Currently we include cost monitoring in the "Observability" capability and [OpenCost](https://github.com/opencost/opencost) in the table. 

For sustainability could you clarify what is the exact capability the platform would provide? Is it the ability to measure and observe power consumption and emissions levels? In that case I'd lean again to say those are part of the observability capability.

Here's another open source project for measuring "sustainability": <https://github.com/sustainable-computing-io/kepler>

In #279 we discussed how capabilities could have many features, so observability would include metrics, traces, costs, power consumption, etc. I wasn't able to add the "features" idea to the paper yet but maybe we should....

PR #312, opened 2023-03-01
[Platforms] Clean refs

joshgav:
This proposes:
- Move list of links from end of whitepaper to separate doc. Let's think about if/how we maintain it going forward - should it be a sort of "awesome list" for platform research?
- Consolidate footnotes and put at bottom of paper.

With this we'll also be able to more easily consider and add other footnotes.


ThomasVitale:
Nice! I like the clean separation between footnotes and resources.

PR #314, opened 2023-03-02
add Kyverno to 'scan' section

chipzoller:
Hi, could I suggest in this PR to add the Kyverno icon to the 'scan' section? Since the definition of of platform here has Kubernetes as a central theme, and many projects herein are exclusively Kubernetes projects, I thought it made sense to represent a CNCF project for policy built exclusively for Kubernetes.


roberthstrand (approved):
Makes sense, and the implementation looks good.

Issue #326, opened 2023-03-07
[Platforms] Get a hosting provider for whitepaper site


joshgav:
In addition to #317 and #318 we also need a hosting provider like [Netlify](https://www.netlify.com/).

joshgav:
Could we use GitHub Pages like we do now for [the TAG blog](https://cncf.github.io/tag-app-delivery/)?

The publishing process for the blog is codified in this branch at the moment: <https://github.com/cncf/tag-app-delivery/tree/jekyll>

joshgav:
After discussions we're going to try to use GitHub pages to host the site as we've been doing for the TAG blog site. We'll add a section for whitepapers in that same site.

Follow ongoing work in #329.

Issue #318, opened 2023-03-03
[Platforms] Get a git repo for whitepaper site


joshgav:
We'll publish the whitepaper as a web site like the cloud-native maturity model here: https://cncfmaturitymodel.netlify.app/, as discussed in #292.

It would be easiest to maintain the site in its own dedicated git repo rather than as part of this repo. This issue is to discuss and track that.

cc @thschue @krumware

krumware:
I think it would make sense just to move the whitepaper to it's own repo, containing the site, as part of the publishing process

joshgav:
Hi @krumware do you think we could have a directory `website` or `site` here in the TAG repo which would host all content for TAG App Delivery including the Platforms and other whitepapers? It could be like Cartografos' [`website` directory](https://github.com/cncf/cartografos/tree/main/website).

At the same time we could move the content currently here in the [Jekyll branch](https://github.com/cncf/tag-app-delivery/tree/jekyll) and published at <https://cncf.github.io/tag-app-delivery/>.

What are the differences between a subdirectory `site` here in this repo versus a dedicated repo just for the site? What would we lose by maintaining the site here? Thanks!

joshgav:
After some discussions we're going to try to use a subdirectory in this repo for the paper; in fact we're going to try to move the TAG blog into this subdirectory too and have the Platforms and other papers as subsections in the greater site.

This implies migrating our TAG blog from Jekyll to Hugo to conform with [Cartografos](https://github.com/cncf/cartografos)' and [Glossary](https://github.com/cncf/glossary)'s patterns. I'll open an issue to track that. See also #327 which brings the `jekyll` branch into `main` - that's not so important now since we'll migrate to Hugo soon.

The [Cartografos](https://github.com/cncf/cartografos) repo uses a subdirectory and the [Glossary](https://github.com/cncf/glossary) repo has site contents in the root. Our plan is to go with Cartografos' approach so we can keep the website separate from other work while benefiting from better discoverability by sharing a single repo.

joshgav:
Work continues to progress in https://github.com/cncf/tag-app-delivery/issues/329

PR #327, opened 2023-03-08
[TAG] Move website content to main branch

joshgav:
This PR proposes merging the website content currently in the [`jekyll` branch](https://github.com/cncf/tag-app-delivery/tree/jekyll) into the main branch in a `website` subdirectory.

This will make it easier for contributors to add content and update styles for our blog - just update the `website` dir in the main branch.

The script isn't needed since the bundled Action "step" executes equivalent functionality, but leaving it just in case.

Verified that the slightly updated workflow runs fine here: <https://github.com/joshgav/tag-app-delivery/actions/runs/4368966252/jobs/7642214572>

Related: #326, #318
cc @krumware @thschue


roberthstrand (approved):

joshgav:
Instead of this we'll move to Hugo and migrate the content as in #333 and #329.

PR #328, opened 2023-03-10
add kubevela as one of the example project in platform whitepaper

wonderflow:
KubeVela is a typical project for platform engineering, it can glue multiple capabilities together and provide an application-centric API. If we can mention this project, the platform whitepaper can be more accurate. Thanks.


roberthstrand (approved):
KubeVela fits the category, and it's a CNCF incubation project. Happy to include it.

PR #332, opened 2023-03-17
initial v2 contribution

AlexsJones:
This contains updated and expanded information about operator comparison, executive review and some other thoughts


joshgav:
Perhaps the directory for updates should be named `latest` rather than `v2`. Otherwise format looks good to me.

joshgav (approved):

thschue (approved):
lgtm

PR #360, opened 2023-03-24
Set up symlinks for whitepapers

cjyabraham:
This implements symlinks for whitepapers since they are not housed within the `/website/` directory.

[Here are the whitepapers on the dev instance](https://deploy-preview-360--tag-app-delivery.netlify.app/whitepapers/)


cjyabraham:
Good idea. I've implemented it, with just symlinks to the `latest` directory for each whitepaper. I think as long as each has an index.md the rest of the assets and files can be called whatever they want.

platforms-whitepaper/v1/index.md
joshgav, line 1: Rather than a new file, can we rename `paper.md` to `index.md` and add the front matter? Thinking to avoid a duplicate which could confuse folks.

joshgav (approved):
LGTM! Thanks @cjyabraham.

platforms-whitepaper/v1/index.md
cjyabraham: I believe that is what I have done? You can see [the directory here for the branch](https://github.com/cncf/tag-app-delivery/tree/symlink/platforms-whitepaper/v1).

joshgav: Oh, okay! I was confused by the GitHub review interface. LGTM.

joshgav:
Thank you @cjyabraham !

PR #365, opened 2023-03-31
[Typo]: as long as long as -> as long as

HyungJune:
as long as long as -> as long as
: I'm not sure that it is the intended sentence. I think it is a typo


thschue (approved):
lgtm

PR #367, opened 2023-04-04
fix: Rename filename to avoid git clone error in Windows

laingke:
An illegal space character at the end of the file name will cause an exception when executing the command `git clone` in Windows.

```
error: invalid path 'gitops-wg/meetings/old-committee-principles/2021-04-28-chat.txt '
error: invalid path 'gitops-wg/meetings/old-monthly/2021-03-11-chat.txt '
error: invalid path 'gitops-wg/meetings/old-monthly/2021-04-15-chat.txt '
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```


joshgav:
Confirming that these files are indeed named with an extra space which should be removed, but this folder requires approval from @scottrigby or @roberthstrand. Can one of you PTAL and merge? Thanks!

roberthstrand (approved):

Issue #378, opened 2023-04-13
CNCF Platforms White Paper - Include KEDA for application autoscaling


tomkerkhove:
I love the CNCF Platforms White Paper, however, I find it a bit of a pity that KEDA (CNCF Incubation, with open proposal for Graduation) is not listed as an application autoscaler.

Is there anything I can help on this front?

PS: As a KEDA maintainer, I'm clearly biased.

niklasbeinghaus:
This issue has been discussed in today's platform-wg meeting.

When abstracting your request for adding KEDA to the platform components, from a broader perspective we would need to have a process for adding new projects first.

Going a bit further, we could also discuss this around the question "Where does _my project_ fit?" Unfortunately, the question isn't exactly easy to answer and we need some more time to discuss how we're gonna proceed.

As a first step, we're likely going to deploy the website from versioned whitepapers folders instead of latest, so you can refer to it as stable artifact: https://github.com/cncf/tag-app-delivery/issues/389

joshgav:
Thanks for the update @niklasbeinghaus and for the suggestion @tomkerkhove. Let's use this issue to discuss where a system like KEDA fits in our current list of capabilities.

I think KEDA and other scaling tools are part of compute infrastructure, and thus the infrastructure capability. I say that cause these tools and similar ones like [Azure VM Scale Sets](https://azure.microsoft.com/en-us/products/virtual-machine-scale-sets/) and [AWS EC2 Auto Scaling](https://aws.amazon.com/ec2/autoscaling/) support adding and removing compute instances as necessary - VMs, containers, functions, perhaps even WASM binaries. The actual scaling of compute is transparent to the developer.

As another example, Knative Serving includes scaling capabilities and perhaps is more clearly part of the compute/infrastructure capability.

These autoscalers don't much effect network and storage either, just compute.

joshgav:
We've merged #385 to add KEDA to the list of example infrastructure capability providers. :tada: 

And we're discussing how to maintain and modify the graphic in #388.

As @niklasbeinghaus says, we still need a way to discuss the current list of "capabilities" and guide projects to find their fit or suggest a new category. I'd like to compare and contrast with the categories in the CNCF landscape and the list of TAGs too. Let's open a new issue to discuss how to modify and maintain the capabilities list.

With that I think it's okay to close this issue, thanks again @tomkerkhove !

tomkerkhove:
No problem for me and thanks for considering and integrating!

[...]

PR #385, opened 2023-04-24
Add KEDA to list of platform components

niklasbeinghaus:
- Infrastructure services
- Messaging and event services

Issue: #378

### DIFF
--- a/platforms-whitepaper/latest/index.md
+++ b/platforms-whitepaper/latest/index.md
@@ -466,7 +466,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>Infrastructure services</td>
     <td>Run application code, connect application components and persist data for applications</td>
-    <td>Kubernetes, Kubevirt, Knative, WasmEdge<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
+    <td>Kubernetes, Kubevirt, Knative, WasmEdge, Keda<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
   </tr>
   <tr>
     <td>Data services</td>
@@ -476,7 +476,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>Messaging and event services</td>
     <td>Enable applications to communicate with each other asynchronously</td>
-    <td>Strimzi, NATS, gRPC, Knative, Dapr</td>
+    <td>Strimzi, NATS, gRPC, Knative, Dapr, Keda</td>
   </tr>
   <tr>
     <td>Identity and secret services</td>
--- a/platforms-whitepaper/latest/index.md
+++ b/platforms-whitepaper/latest/index.md
@@ -466,7 +466,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>Infrastructure services</td>
     <td>Run application code, connect application components and persist data for applications</td>
-    <td>Kubernetes, Kubevirt, Knative, WasmEdge, Keda<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
+    <td>Kubernetes, Kubevirt, Knative, WasmEdge, KEDA<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
   </tr>
   <tr>
     <td>Data services</td>
--- a/platforms-whitepaper/latest/index.md
+++ b/platforms-whitepaper/latest/index.md
@@ -476,7 +476,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>Messaging and event services</td>
     <td>Enable applications to communicate with each other asynchronously</td>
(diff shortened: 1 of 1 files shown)


platforms-whitepaper/latest/index.md
```diff
@@ -466,7 +466,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>Infrastructure services</td>
     <td>Run application code, connect application components and persist data for applications</td>
-    <td>Kubernetes, Kubevirt, Knative, WasmEdge<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
+    <td>Kubernetes, Kubevirt, Knative, WasmEdge, Keda<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
```
tomkerkhove, line 469:
```suggestion
    <td>Kubernetes, Kubevirt, Knative, WasmEdge, KEDA<br />CNI, Istio, Cilium, Envoy, Linkerd, CoreDNS<br />Rook, Longhorn, Etcd</td>
```

platforms-whitepaper/latest/index.md
```diff
@@ -476,7 +476,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>Messaging and event services</td>
     <td>Enable applications to communicate with each other asynchronously</td>
-    <td>Strimzi, NATS, gRPC, Knative, Dapr</td>
+    <td>Strimzi, NATS, gRPC, Knative, Dapr, Keda</td>
```
tomkerkhove, line 479:
```suggestion
    <td>Strimzi, NATS, gRPC, Knative, Dapr, KEDA</td>
```

tomkerkhove (changes requested):
Sorry for the massive nits but we have to follow KEDA's branding guidelines - https://github.com/kedacore/governance/blob/main/BRANDING.md

tomkerkhove:
Sorry to be a pain, but wouldn't it be nice if we put autoscaling front and center by introducing a new category for it? Scalability is at the heart of cloud-native so I believe we should emphasize this here.

Not only to cover KEDA, but potentially other things such as Kubernetes Cluster Autoscaler or Karpenter.

Also, can we update this visual overview?

image

niklasbeinghaus:
> Sorry to be a pain, but wouldn't it be nice if we put autoscaling front and center by introducing a new category for it? Scalability is at the heart of cloud-native so I believe we should emphasize this here.

I wasn't quite sure about the categories in this either, however from a platform engineering perspective, it makes sense to keep this condensed. What do you think about splitting the "automate" category and adding a "scaling" category thereunder?

> Also, can we update this visual overview?
Once we've found a solution for the right category, we should definitely update the visual overview as well.

tomkerkhove:
> I wasn't quite sure about the categories in this either, however from a platform engineering perspective, it makes sense to keep this condensed. What do you think about splitting the "automate" category and adding a "scaling" category thereunder?

Fine for me!

joshgav:
Hi @tomkerkhove, I lean to think (auto)scaling is a feature of whatever resource is being managed. Since KEDA scales compute infrastructure based on events it seems it's a feature of compute.

Karpenter and Cluster Autoscaler are interesting examples. Still, they're mostly focused on automatically scaling compute infrastructure and so probably belong with "Infrastructure". Do they also automate scaling of other resource types?

So I'd say for now let's put autoscalers together with the resource types they scale, in KEDA's case infrastructure/compute.

joshgav (changes requested):
We've tried to list any given project only once to reduce clutter. Though it fits a bit under "events" too, I think the primary place for KEDA would be within infrastructure since it manages compute. Could you put it just in the infrastructure category? Thanks!

zroubalik:
> Hi @tomkerkhove, I lean to think (auto)scaling is a feature of whatever resource is being managed. Since KEDA scales compute infrastructure based on events it seems it's a feature of compute.
> 

Hey, it depends on what do you consider as `compute infrastructure`, KEDA scales workloads (apps - Deployments, Jobs...).

> Karpenter and Cluster Autoscaler are interesting examples. Still, they're mostly focused on automatically scaling compute infrastructure and so probably belong with "Infrastructure". Do they also automate scaling of other resource types?

These tools scales only compute nodes (k8s nodes) and if I am not mistaken Karpenter is AWS only.

abangser:
I would agree that this still feels like compute scaling. IMO deployments / jobs / CRDs are versions of compute, just like bare metal, VMs, and containers are versions of compute. We are getting better at abstractions, but at the end of the day my understanding is that KEDA helps manage higher load on the system by horizontally scaling the things that respond to those higher requests. Would be really helpful to understand if that is the wrong understanding though!

zroubalik:
> I would agree that this still feels like compute scaling. IMO deployments / jobs / CRDs are versions of compute, just like bare metal, VMs, and containers are versions of compute. We are getting better at abstractions, but at the end of the day my understanding is that KEDA helps manage higher load on the system by horizontally scaling the things that respond to those higher requests. Would be really helpful to understand if that is the wrong understanding though!

Yeah, that's completely correct understanding! As I mentioned, it depends on what do you see by `compute`, if it is what you just mentioned above, then we should be good.

joshgav (approved):
LGTM! Thanks for the productive discussion!

PR #413, opened 2023-06-20
Platform White-paper - adding cdk8s to the CNCF projects table

tsahiduek:
I think that the [Cloud Development Kit for Kubernetes (cdk8s)](https://cdk8s.io/) project is another example of "APIs for automatically provisioning capabilities" tooling.

### DIFF
--- a/platforms-whitepaper/latest/index.md
+++ b/platforms-whitepaper/latest/index.md
@@ -436,7 +436,7 @@ relating it to existing CNCF or CDF projects.
   <tr>
     <td>APIs for automatically provisioning capabilities</td>
     <td>Structured formats for automatically creating, updating, deleting and observing capabilities.</td>
-    <td>Kubernetes, Crossplane, Operator Framework, Helm, KubeVela</td>
+    <td>Kubernetes, Crossplane, Operator Framework, Helm, KubeVela, cdk8s</td>
   </tr>
   <tr>
     <td>Golden path templates and docs</td>


abangser:
Thanks so much @tsahiduek for suggesting CDK8s. I think it is a fantastic project! I just wonder your thinking on the location on the table which I think actually drives a bigger question about refactoring the white paper table. Therefore, I am going to split my comment into two:

1. I agree and endorse adding CDK8s to the white paper list of projects. It is a part of the CNCF (sandbox) and is a great project for generating tooling for platforms. For that reason, I think we should merge this PR 🚀

2. I want to use this PR to see if we should open a separate conversation / PR to revisit what we mean by `APIs for automatically provisioning capabilities` as a whole. IMO the current group of tools breaks down into two categories:
    1. _Tools that can provide an API that separates implementation from consumption_: Kubernetes, Crossplane, Operator Framework
    2. _Tools that provide more readable and template-able code when creating resources_: Helm, KubeVela, cdk8s 

    I believe that to clarify the push from DevOps or dev enablement towards platform engineering, we need to be evangelising the focus on decoupling the producers and consumers of services. IMO there will always be dev enablement that makes the consuming of services easier. But with platform engineering we are moving the value line of what should be provided as a service inside a company. We are now naming (and therefore formalising) the expectation that internal company teams should be providing company specific business processes with as much polish/servicing as we have come to expect from a IaaS product like Amazon EC2 or Google Cloud Run or Azure Blob Store (or add your IaaS offering of choice here!).

So to make this actionable and so as not to stick up this PR. What I am asking is that reviewers look to review/approve/merge this PR. And while here, if you feel compelled to speak further on my 2nd point, please react to this post and and I will be happy to open a new issue to make sure this point is sufficiently discussed. More clearly, 😕/👎 can indicate confusion or disagreement, and 🚀 /👍 can indicate agreement, but any of these emojis will result in the action of opening an issue!

joshgav:
We are proposing replacing project logos with generic graphics, please see #471. With that we won't need to determine the placement of projects in categories for this graphic. Can you please take a look at #471 and approve?

PR #426, opened 2023-07-20
Website restructuring

lianmakesthings:
Draft for new website as discussed in this issue https://github.com/cncf/tag-app-delivery/issues/410
Discussion Doc: https://docs.google.com/document/d/1RibiFs3KlQ0FhfaF4X8owQmDzLuN2y4jRnQy3TYp6KI/edit

PR #450, opened 2023-08-16
Enable  i18n with zh(Chinese) for Platforms paper

node:
Use the latest website version (after #426 ).


abangser:
Hey @node , thanks for this PR! I am really excited to see us set this site up for multiple languages and specifically for the white paper 🎉

I used your preview link from [this comment](https://github.com/cncf/tag-app-delivery/pull/450#issuecomment-1680379495) and tried to navigate to the white paper and saw some behaviour I didn't expect.

Specifically, I followed this link path:
1. Open the home page, note that the language in the top right is `English`
2. Click on `Working Groups` in the top right
3. Click on `Platform Working Group` (recreated from both the main body of the page or the left hand side link)
4. Click `CNCF Platforms White Paper` from the left hand side

When this page loaded I saw a few things:
1. The language in the top right turned to `Chinese`
2. The language in the whole page was a mix (see attached screenshot)
3. I could not change the language back to `English`

This seems to be an issue only on the white paper page since the charter and other pages seem to allow both languages.

Can you please take a look and make the white paper experience the same as the other pages before we look to review / merge? Thanks!

node:
Hi @abangser  Thanks for your review and bring up the issue.  

I checked the config and found the key issue: there is a soft link under both content/en/wgs/platforms  and contene/zh/wgs/platforms , and they refer the same one file under platforms-whitepaper/latest . So we can not let this support i18n, better choice is copy it then translate . 

I will fix this later  😄

node:
Hi @abangser   The PR updated has fixed the issue.  The key point is the URL attribute setting in .MD file under whitepaper. 
Cancle that setting, everything is OK now.

abangser:
Amazing, thanks for the quick review @node 🙌 

Just want to confirm that this is to put in the drop down for language in the top right plus minimal title translations rather than any full document translations?

Assuming that is correct, I think that this is what I would expect and all looks good. I believe some of the people with repo permissions for merging may be on summer holiday so I apologise if this sits short bit, but I will attempt to chase as they come back.

node:
Correctly .  Only main menu and page/article titles are translated into Chinese.

By the way, as I know the platform whitepaper have been translated into Chinese by other guy, maybe we cloud suggest the translator submit here again after this PR merged.

roberthstrand (approved):
I looked through the preview deployment, and everything looks good. No broken links, that I can find at least.

website/content/en/wgs/operator/whitepaper/README.md
joshgav, line 1: This directory should be a symlink to the top-level "operator-whitepaper/latest" directory.

joshgav (changes requested):
Thanks @node for this work! Looks pretty good to me, except that originally the whitepaper directories were symlinks to the directories in the repo `operator-whitepaper/latest` and `platforms-whitepaper/latest`. Can we restore those symlinks for the English version?

Side note - I'm open to revisiting the symlink setup, if we want to do that let's open another issue to discuss.

Thanks again! And indeed, we'll ask the contributor of the Chinese translation to submit it here once this is merged.

website/content/en/wgs/operator/whitepaper/README.md
abangser: Did we have latest? I thought we left it as vX so that it would stay stable for multiple languages as we added / evolved latest. 

That was my memory but also remember this being an interesting debate so please excuse me if I misremember!

joshgav: We did talk about that @abangser and could visit again, but it looks like we didn't implement it or it got reverted. Definitely open to making the change in another PR. My ask is to stick with the current arrangement for this PR; here's the symlink now: <https://github.com/cncf/tag-app-delivery/blob/main/website/content/wgs/platforms/whitepaper>

An interesting question this raises - should the whitepaper directories themselves be split into languages? Where will the official Chinese translation go - at the moment would it only be in the website directory?

abangser: Definitely happy to stay latest for this PR 👍

But also great shout on having confidence on where the translation files will go as a part of merging this one.

website/content/en/wgs/platforms/whitepaper/index.md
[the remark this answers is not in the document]
node: Restore symlinks of two whitepapers under content/en .

website/content/en/wgs/operator/whitepaper/README.md
node: That is really hard to keep latest version in multi languages without outer translation track system or tool. 

I think currently we choose english version as latest util there is a final version or relatively stable version, then sync to other language content dir. 

So I restore the symlinks under content/en and keep a copy version under content/zh .

joshgav:
@node we still want to include these, I've just been waiting for some of our backlog to wrap up.

node:
> @node we still want to include these, I've just been waiting for some of our backlog to wrap up.

There are some conflicts in the previous PR, I will check and fix them then re-submit  a new PR.

node:
@joshgav  Make it this time, hard but successfully 😄      See #492

joshgav:
Thanks you @node! Taking a look.

PR #459, opened 2023-09-15
Update _index.md with Chairs and TLs

angellk:


joshgav (approved):
LGTM!

We also need to update the CODEOWNERS file.

thschue (approved):
lgtm

PR #463, opened 2023-09-26
Extract glossary from Platforms White Paper to top level concern

abangser:
In writing the platforms white paper it was realised that glossary would help clarify some key terms. These were originally added at the end of the white paper for proximity and ease.

During the review of the upcoming platform engineering maturity model it was realised that the entries in this glossary would be useful to the maturity model readers as well.

This PR extracts the glossary into its own top level content. It intentionally does not introduce any new terms or improve any existing terms as that work should be evaluated in independent PRs.

The only "decisions" made in this PR are:
1. The glossary deserves to be its own stand alone page
2. The glossary should be a ~TAG level concern (despite it currently only containing WG Platform terms).~ WG level concern (per feedback in this PR).

image

### DIFF
--- a/platforms-wg/glossary/_index.md
+++ b/platforms-wg/glossary/_index.md
@@ -1,10 +1,6 @@
 ---
 title: Glossary
-list_pages: true
-menu:
-  main:
-    weight: 30
-description: "A collection of terms and used within the TAG published papers."
+description: "A collection of terms used in the Platform Working Group published papers."
 ---
 
 See also <https://glossary.cncf.io/>.
--- a/platforms-whitepaper/latest/index.md
+++ b/platforms-whitepaper/latest/index.md
@@ -497,52 +497,6 @@ relating it to existing CNCF or CDF projects.
   </tr>
 </table>
 
-## Glossary
-
-See also <https://glossary.cncf.io/>.
-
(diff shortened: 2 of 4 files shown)


joshgav:
Thanks for this @abangser, agree with the change as we discussed in the meeting. Can we put the website content in  `website/content/wgs/platforms`? We can put the maturity model there too.

Also, should we keep the canonical copy in the wg-platforms directory? And we can symlink that canonical copy into the "WG Platforms" directory, i.e. `website/content/wgs/platforms`.

Thanks!

abangser:
Thanks for reviewing @joshgav!

If I understand correctly you are suggesting this live under the WG and I am happy with this as I can see an argument for either TAG level or WG level. Given current contents being 100% platform WG related it makes sense to start there 👍

I have added a new commit (in case I misunderstood your feedback), updated the PR details to accurately depict the decisions made, and added a screenshot for easier review 🙌 If this is good, feel free to squash these as the two commits purely give us the chance to go back to a TAG level glossary if I misunderstood!

joshgav:
Yes, thank you Abby, and thanks for the screenshot! Wish the auto-render was working (cc @cjyabraham).

LGTM!

joshgav (approved):
LGTM, thanks @abangser!

PR #478, opened 2023-10-21
move opening content into context tab

joshgav:
This is a proposal to consolidate the introduction section so readers get to the model faster. The idea is to make the document easier to consume at a glance by moving some of the context to a tab for later consideration.

PTAL @abangser, WDYT?

Here's a rendered version: <https://maturity-model-v1a--frolicking-mandazi-6b5fd4.netlify.app/whitepapers/platform-eng-maturity-model/>

This PR:

- Consolidates the introduction section and moves some context about the model into a 6th tab in the Model Details section named "Context"
- Tweaks some markdown codes for more distinct headings
- Uses `padding-top:12px;padding-bottom:4px` for header styles instead of extra `<br />`'s where feasible
- Reduces conclusion paragraph to one sentence, relying more on the Context tab

### DIFF
--- a/platforms-maturity-model/v1/index.md
+++ b/platforms-maturity-model/v1/index.md
@@ -12,7 +12,7 @@ url: /whitepapers/platform-eng-maturity-model
 
 ## Introduction
 
-CNCF's [Platforms Definition paper](https://tag-app-delivery.cncf.io/whitepapers/platforms/) describes what internal platforms for cloud computing are and the values they promise to deliver to enterprises. But to achieve those values an organization must reflect and deliberately pursue outcomes and practices that are impactful for them, keeping in mind that each platform is custom-crafted and maintained for its own organization. This maturity model provides a framework for that reflection and for identifying opportunities for improving platforms in any organization.
+CNCF's initial [Platforms Definition white paper](https://tag-app-delivery.cncf.io/whitepapers/platforms/) describes what internal platforms for cloud computing are and the values they promise to deliver to enterprises. But to achieve those values an organization must reflect and deliberately pursue outcomes and practices that are impactful for them, keeping in mind that every organisation relies on an internal platform crafted for its own organization even if it is just documentation on how to use third party services. This maturity model provides a framework for that reflection and for identifying opportunities for improvement in any organization.
 
 ## What is platform engineering?
 
@@ -26,7 +26,7 @@ Please read the [CNCF Platforms Definition white paper](https://tag-app-delivery
 
 As platform engineering has risen in prominence over the last few years, some patterns have become apparent. By organizing those patterns and observations into a progressive maturity model, we aim to orient [platform teams]({{< ref "/wgs/platforms/glossary#platform-teams" >}}) to the challenges they may face and opportunities to aim for. Each aspect is described by a continuum of characteristics of different teams and organizations at each level within the aspect.
 
-Each additional level of maturity is accompanied by greater requirements for funding and people's time. Therefore, reaching the highest level should not be a goal. Each level describes qualities that should appear at that stage. Readers must consider if their organization and their current context would benefit from these qualities given the required investment.
+Each additional level of maturity is accompanied by greater requirements for funding and people's time. Therefore, reaching the highest level should not be a goal in itself. Each level describes qualities that should appear at that stage. Readers must consider if their organization and their current context would benefit from these qualities given the required investment.
 
 Keep in mind that each aspect is meant to be evaluated and evolved independently. However, as in any socio-technical system these aspects are complex and interrelated. Thus you may find that to improve in one aspect you must reach a minimum level in another aspect too.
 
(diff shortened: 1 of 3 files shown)


platforms-maturity-model/v1/index.md
```diff
@@ -1,85 +1,38 @@
...
+Inspired by the cross-functional cooperation promised by DevOps, platforms and platform engineering have emerged in enterprises as an explicit form of that cooperation. Platforms curate and present common capabilities, frameworks and experiences that facilitate and accelerate the work of [internal users]({{< ref "/wgs/platforms/glossary#platform-users" >}}) such as product and application teams.
 
-For example, if your organization is large enough to benefit from scaling platform use to many product teams, then you will probably find it useful to grow and invest in your platform like you do customer-facing products. You will probably consider levels 1 and 2 to be steps in your organization's journey — opportunities to research and develop solutions and prepare for later levels. Levels 3 and 4 would likely be your desired ultimate state for your platforms, when you'll be prepared to scale to broad use while sustainably updating and maintaining them.
```
abangser, line 50: I noticed this has been removed in this PR though it is hard to spot given the diff.

Would it be possible for you to call out any major shifts in the content along with the change in the location of the content? I had started doing a review on the v1 branch but am realising that won't be helpful as this one will make some updates!

platforms-maturity-model/v1/index.md
```diff
@@ -497,19 +436,49 @@ Not only are standard frameworks leveraged, but there is an understanding that m
...
+<h4 style="color:gray;padding-bottom:4px;padding-top:12px"><i>Context of this work</i></h4>
 
-</br>
-</br>
+It's valuable to understand the context a document has been written in. The following sections lay out some context behind the model as well as some expectations for you, the reader.
 
----
+## Why provide a maturity model?
+
+This model encourages organizations to mature both their platform engineering discipline and the resulting platform through intentional planning. This effective planning and discipline is itself a precursor to effective platform evolution as organizational requirements change.
```
abangser, line 447: IMO this is an interesting new section title but the content is not quite fit for purpose. If we want to tackle the decision to settle on a maturity model I think we need to ack the challenges these models have had in the past (consultant fodder) and why we still felt the structure provided the most possible value (narrowing focus, providing guideposts, acknowledging when behaviours result in less effective solutions etc).

I took a stab at it but am not happy with my version so decided it would be best to call this out and decide if we really want this section.

platforms-maturity-model/v1/index.md
joshgav: That makes sense, thanks for catching this @abangser. I think this may be the one big chunk I dropped but I'll go back through again. Do you think we should add it back somewhere?

platforms-maturity-model/v1/index.md
joshgav: Yeah, it felt like it might be a question folks ask but agree the paragraph is only a partial answer. Thinking about it.

joshgav:
@abangser merged in your recommended changes. Remaining is to be sure no other significant chunks were excluded (I don't think they were) and to settle on how to handle the "Why a maturity model?" heading - I think we could find a narrower heading, thinking about it.

joshgav:
Here are the 3 paragraphs which this PR proposes removing.

These first 2 seem to be well covered by the rest of the content and so seem okay to remove in the interest of being concise, but if folks think otherwise we could integrate them back in the "Context" tab.

> As an organization learns and evolves, so will its platform practices mature and develop. However, an organization should not strive for “Level 4” maturity “just because.” Instead, each organization should evaluate its own particular requirements and invest accordingly in platform capabilities that are most appropriate for its needs. Said another way, this model is not a ruler by which the success for any given platform, person, team, or organization should be measured. It is instead offered as a tool for users to identify and articulate opportunities for enhancement and to solicit support to optimize software delivery and platform practices. 

> For example, if your organization is large enough to benefit from scaling platform use to many product teams, then you will probably find it useful to grow and invest in your platform like you do customer-facing products. You will probably consider levels 1 and 2 to be steps in your organization’s journey — opportunities to research and develop solutions and prepare for later levels. Levels 3 and 4 would likely be your desired ultimate state for your platforms, when you’ll be prepared to scale to broad use while sustainably updating and maintaining them. In contrast, if your organization is just a handful of engineers working on an early product prototype, it may be best to rely on one-off processes and ask product teams to own any platform capabilities they require.

This last one doesn't seem that accurate or important frankly - and I've been the one wanting it :stuck_out_tongue_winking_eye:. I do think there's opportunity to backport some of the ideas in the maturity model back to the definitions paper at some point but in the meantime I think we should leave it out.

> The top-level Platforms white paper introduces a number of [attributes that define successful platforms](https://tag-app-delivery.cncf.io/whitepapers/platforms/#attributes-of-platforms); and the aspects of this maturity model are inspired by and correlated with those attributes.

platforms-maturity-model/v1/index.md
joshgav: In the latest commit I moved this paragraph up into the top section and removed this heading entirely, see what you think.

PR #479, opened 2023-10-22
cleanup home page

joshgav:
- fixes links to lead names
- makes text in header a bit more concise
- makes text in body a bit more descriptive
- moves meetings info up right after descriptions
- moves picture to the bottom


lianmakesthings (approved):
lgtm

PR #472, opened 2023-10-16
Platform Engineering Maturity Model v1

joshgav:
This is the culmination of WG Platforms' last few months of effort to provide a next step for readers of the Platforms White Paper. We'll merge this to main to finalize publication and represent final acceptance by the TAG and CNCF TOC.

PR #482, opened 2023-10-25
Maturity model v1

abangser:
This change has 4 commits, each for a different value:
1. Fix the links from the table to the tabs
2. Move the context up from in the tabs to the body of the paper
3. Fix the horizontal bar delineating between the table details and the conclusion
4. Nit wording change on one sub title

These changes can be evaluated for merge separately and if needed I can break them apart. The real core of this is around where the context for the model should live. This change is being proposed based on @krumware's ask [here](https://github.com/cncf/tag-app-delivery/pull/472#issuecomment-1776333603) which was discussed on the WG call on Tuesday Oct 24. It seemed like there was strong support for making the context more central to the doc by moving it back and that this would actually reduce confusion caused by the multiple uses of the tabs.

cc/ @joshgav

### DIFF
--- a/platforms-maturity-model/v1/index.md
+++ b/platforms-maturity-model/v1/index.md
@@ -80,7 +80,7 @@ This yields a potential process for platform development and maturity similar to
 
 Similar to the [CNCF Cloud Native Maturity Model](https://maturitymodel.cncf.io/), this model highlights that successful business outcomes can only be achieved through balancing people, process, and policy alongside technology. Notably, this model introduces aspects which are often not fully in the remit of a single internal team, but rather require cooperation across the engineering department and quite often the wider organization.
 
-### But we don’t fit
+### But it doesn't seem to fit
 
 That’s perfectly fine! All organizations and groups have dynamics and parameters that are specific to them.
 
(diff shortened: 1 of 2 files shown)


platforms-maturity-model/v1/index.md
```diff
@@ -66,26 +80,29 @@ This yields a potential process for platform development and maturity similar to
...
+| [Adoption](#Adoption)         | _Why and how do users discover and use internal platforms and platform capabilities?_      | Erratic                | Extrinsic push        | Intrinsic pull         | Participatory                |
+| [Interfaces](#Interfaces)     | _How do users interact with and consume platform capabilities?_                            | Custom processes       | Standard tooling      | Self-service solutions | Integrated services          |
+| [Operations](#Operations)     | _How are platforms and their capabilities planned, prioritized, developed and maintained?_ | By request             | Centrally tracked     | Centrally enabled      | Managed services             |
+| [Measurement](#Measurement)   | _What is the process for gathering and incorporating feedback and learning?_               | Ad hoc                 | Consistent collection | Insights               | Quantitative and qualitative |
 
-{{< /tab >}}
+## Model Detail
+
+<div style="min-width:620px">
+{{< tabs tabTotal="6">}}
```
joshgav, line 105: Does this need to be 5 now?

joshgav (approved):
LGTM!

joshgav:
Rendered: https://deploy-preview-472--tag-app-delivery.netlify.app/whitepapers/platform-eng-maturity-model/

platforms-maturity-model/v1/index.md
abangser: Ah I think it does 🙈

PR #487, opened 2023-10-31
update WG directory pattern in CODEOWNERS

### LINKED ISSUE #287: [TAG] Fix up CODEOWNERS
This has been bugging me for some time, but I haven't gotten around to raising the issue. There are several users set as codeowner that does not have any permissions on the repository. There are a couple of users here that might be stale, so I can figure out who should have access, but I need help from an owner to grant permissions to users.

joshgav:
Updates CODEOWNERS so that WG's own all of their directories.

Also notes reason for error in GitHub: it's because named users are not members of the repository yet. In the current case their invites had expired so I resent.

fixes #287

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,5 @@
+# NOTE: The error in GitHub is for users that have not yet accepted invites to this repository; it is expected.
+
 # CODEOWNERS file indicates code owners for certain files
 #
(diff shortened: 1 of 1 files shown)


roberthstrand (approved):

lianmakesthings (approved):
lgtm

PR #488, opened 2023-10-31
Update Kubecon Chicago schedule

joshgav:
Adds more presentations to the schedule for Monday and Wednesday.


lianmakesthings (approved):
Awesome to see slots filling up! 👏

PR #492, opened 2023-11-02
Enable i18n with zh(Chinese) for our new website

node:
- [ ] Enable i18n by update website/config.toml  and website/i18n/*.toml 
- [ ] move content/* under content/en/, update the symlinks  
- [ ] copy all contents to content/zh  incloud symlinks files 
- [ ] test locally - Everything is OK

See the previous closed PR #450  for more history messages.


node:
Update: 
The translation of basic information on the homepage has been updated, as well as the minimal translation of other menus, without involving specific content translation at the moment.

abangser (commented):
Very excited to see this get merged in! It all seems to navigate OK for me 🎉

One worry I have is about the sustainability of blogs (or any other content) being in multiple places even if not translated. Is there a way to do a fall back to the baseline language (english) if the translation does not exist instead of requiring a copy in each language directory?

node:
> One worry I have is about the sustainability of blogs (or any other content) being in multiple places even if not translated. Is there a way to do a fall back to the baseline language (english) if the translation does not exist instead of requiring a copy in each language directory?

That is possible. We can add only one article with a link to English blog under blog menu, and guide users to English version to visit more.

joshgav:
I like Abby's suggestion of simply falling back to English content if the desired language isn't available - is that possible in Hugo? (cc @cjyabraham do you know?)

I'm less a fan of not having the blog in Chinese or other languages at all - I lean to say let's copy the content for now and have a Chinese blog, this allows for native Chinese blog posts too.

Thanks @node for getting this started!

cjyabraham:
> I like Abby's suggestion of simply falling back to English content if the desired language isn't available - is that possible in Hugo? (cc @cjyabraham do you know?)

I think that might be a strange UX. It would mean someone who deliberately chooses Chinese goes instead to the English page.

Currently if a translation of a page is missing and someone chooses Chinese, it takes them to the homepage of the site in Chinese, which I think makes more sense. At that point the user can see in the nav what pages are available in that language.

I'm observing on the Glossary as well that normally if a page isn't offered in a particular translation, then that language isn't listed in the language dropdown, which makes sense. For whatever reason that doesn't seem to function when there's only the one other language.

kgamanji:
Great initiative - definitely support the multi-language initiative for the TAG App-Delivery website.

node:
Thanks for your review. Regarding the blog, it is suggested that we put aside the controversy or continue to discuss better solutions for translating or not translating the blog in a new issue. Apart from the blog, if everything is normal, I suggest that we move forward with the launch as soon as possible so that those who are interested in Chinese translation can participate early on. In fact, as of now, I have seen at least two developers complete the translation of the platform white paper and platform engineering maturity model and are looking for a suitable submission entry.

joshgav:
Thank you @node! I merged and let's move forward from here. Thank you for this contribution!

PR #493, opened 2023-11-05
Fix ignore command for builds

cjyabraham:
According to [this thread](https://answers.netlify.com/t/support-guide-how-to-use-the-ignore-command/37517) I should be using the `exit` command here.


joshgav (approved):
Makes sense to me, thanks @cjyabraham !

PR #505, opened 2023-11-19
See if this builds a preview

cjyabraham:


cjyabraham:
The build preview was created on this PR. Strange since it was not created on https://github.com/cncf/tag-app-delivery/pull/502.

Issue #506, opened 2023-11-21
Do we want feedback buttons on pages of the site?


cjyabraham:
I've recently got the page feedback buttons to work on the [CNCF Glossary site](https://glossary.cncf.io/abstraction/) (see at the bottom of the page). They feed events into our cncf.io Google Analytics 4 property which we can later create reports on to see which pages get good feedback, which do not.

<img src="https://github.com/cncf/cartografos/assets/463160/71e21b1f-90ad-44a3-acca-063619fb4adc" width=400>

Would we like this feature activating for the TAG App Delivery site? If not, please just close this issue.

cjyabraham:
Any thoughts @joshgav ?

joshgav:
Hi @cjyabraham, we certainly love feedback, so sure let's add these!

It would really help us improve if we get reports with this feedback and information on traffic patterns and user behavior. Maybe you could share all this with us once per quarter or so @cjyabraham, WDYT?

cjyabraham:
Sounds good. I've submitted a PR: https://github.com/cncf/tag-app-delivery/pull/510

cjyabraham:
This is deployed. In a couple of weeks I'll produce a report to see if there's any useful data.

PR #510, opened 2023-11-23
Add feedback buttons to pages

cjyabraham:
Addresses #506

### DIFF
--- a/website/config.toml
+++ b/website/config.toml
@@ -140,8 +140,8 @@ sidebar_search_disable = false
 [params.ui.feedback]
 enable = true
 # The responses that the user sees after clicking "yes" (the page was helpful) or "no" (the page was not helpful).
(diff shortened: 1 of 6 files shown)


abangser:
Confirmed I was able to click one of the feedback buttons on the white paper page 🙌

Not sure which pages should/should not have them so will leave that for the lead review, but very happy to see this implemented! 🎉

joshgav (approved):
LGTM.

@cjyabraham would you mind listing exactly what page types we should expect to have the feedback button? Is it blog pages and whitepaper pages?

cjyabraham:
It's currently set to show the buttons on [blog posts](https://tag-app-delivery.cncf.io/blog/infrastructure-for-apps-platforms-for-cooperative-delivery/), contribute section pages (of which there currently aren't any), and [whitepaper pages](https://tag-app-delivery.cncf.io/whitepapers/operator/). Note that it won't show on any list pages which are those that have the filename `_index.md`. I think some of those actually should just be regular pages, with filename `index.md`.

PR #515, opened 2023-12-07
Blog post: Platform Engineering Through the Lens of Two-Sided Markets

nagyv:
Closes https://github.com/cncf/tag-app-delivery/issues/501

Issue #495, opened 2023-11-06
Blog post on how to get started with contributing to the TAG app delivery - WG Platforms


techmaharaj:
Idea: Create a blog post about how can people get started on contributing to the TAG app delivery. Largely on the line of creating an issue, sharing the idea, joining the meetings, discussing the outcomes, TG, etc., and eventually publishing the blog post.

menzenski:
I think there's potential also to tie into the existing Platforms Working Group Charter, specifically the part of that document which discusses the working mode of the group: https://tag-app-delivery.cncf.io/wgs/platforms/charter/#working-mode--expected-outcome .

abangser:
Agreed @menzenski I am going to take a stab at some work to codify a process and I know @adminturneddevops is interested in this too. 

I will look to first create a Google doc and then a PR with the content and will comment on this issue as the work progresses for feedback/transparency as I see this issue as the culmination of that work. 

My target timeframe is an MVP ready in days/weeks rather than a highly structured thing that takes weeks/months to get rolled out 🤞

AdminTurnedDevOps:
Something that I was thinking here are some basic steps and then dive into it more in-depth. For example, I think a great first step is to read through the CNCF WG Slack channel that you think you may be interested in and ask the simple question of "hey, anywhere I can help out?". We could go more in-depth as well and say something like "here are some common questions you can ask to get yourself started"

abangser:
Great idea @AdminTurnedDevOps, I think that is maybe another angle on the doc I just wrote [here](https://docs.google.com/document/d/1HyjXbR2VJ5Pj6qn7wASvEY5tH7YhTAVm96Y4BuOv1NE/edit?usp=sharing). As in, some people come with proposals and some just come to interact and we need to support both.

Can you take a look at my suggestions around supporting early contributors proposals and tell me what you think? Please keep in your mind how it felt to come with your idea around the tech stacks and what we can/should do to make that easier for you now and future yous!

AdminTurnedDevOps:
@abangser Awesome, thanks! I added in some suggestions and comments (hope that's okay). Let me know if you have any questions. Happy to help!

techmaharaj:
@abangser I had a look at this too and left one suggestion on the steps, the rest of it looks pretty good. 

Further, in my head, I would read the current headings as "Who we are?", "What we do?" and "How you can get started?" - only to keep it simple and because these _might have better relatability_ for first-timers as compared to the current titles. WDYT?

[...]

PR #518, opened 2023-12-16
Introduce taxonomy for blog pages

abangser:
Hugo has the built in ability to do categories and tags. Using them can enable things like word clouds, pages that highlight aligned content and more.

Right now this is just the first introduction of categories and tags and is being used only for styling a single blog post. This formatting will help us enable more community posts by being clear when things are TAG/WG created vs community submitted.

This will be formalised in #495 and be retrospectively applied to #515 and available for a few more that I know people are interested in proposing.

The proposed taxonomy is:

* *Categories* should be minimal and applicable across a number of tags. Therefore there are only two:
  * announcements
  * community
* *Tags* can help identify sub topics that are more narrow focuses. Right now there are proposed tags:
  * KubeCon
  * Technology
  * WG Multi-tenancy
  * WG Platforms

## Help wanted

I am no designer, to say the least 😅 This could use a look over on they use of Hugo (maybe @cjyabraham ?) and also general desired design (also maybe Chris, but also anyone else!)

### DIFF
--- a/website/content/en/blog/clusters-for-all-cloud-tenants.md
+++ b/website/content/en/blog/clusters-for-all-cloud-tenants.md
@@ -1,7 +1,12 @@
 ---
 title:  "Clusters for all cloud tenants"
 date:   2022-06-02 13:04:00 +0200
-author: Josh Gavant 
+author: Josh Gavant
+categories:
+- Article
+tags:
+- WG Multi-tenancy
+- Community Contribution
 ---
 
 A decision which faces many large organizations as they adopt cloud architecture is how to provide isolated spaces within the same environments and clusters for various teams and purposes. For example, marketing and sales applications may need to be isolated from an organization's customer-facing applications; and development teams building any app usually require extra spaces for tests and verification.
--- a/website/content/en/_index.md
+++ b/website/content/en/_index.md
@@ -20,7 +20,7 @@ The TAG gathers feedback from cloud-native application developers, platform
 engineers and end users; shares that feedback with projects in its domain; and
 produces guidance and examples for end users.
 
-The TAG supports projects relatated to its charter
+The TAG supports projects related to its charter
 such as those in the [CNCF Landscape](https://landscape.cncf.io/card-mode) under
 [application definition and image build](https://landscape.cncf.io/card-mode?category=application-definition-image-build&project=hosted),
 [continuous integration and delivery](https://landscape.cncf.io/card-mode?category=continuous-integration-delivery&project=hosted)
@@ -58,4 +58,3 @@ as well as the [CNCF Community Calendar](https://community.cncf.io/tag-app-deliv
 - [Mailing list](https://lists.cncf.io/g/cncf-tag-app-delivery/topics)
 
 <p class="mt-5"><img src="/images/man-using-laptop.jpg" alt="Man working on computer"></p>
-
--- a/website/content/en/blog/2023-10-kubecon-chicago.md
+++ b/website/content/en/blog/2023-10-kubecon-chicago.md
@@ -3,6 +3,10 @@ title:  TAG App Delivery at Kubecon Chicago
 slug:   tag-app-delivery-at-kubecon-chicago
 date:   2023-10-09 12:00:00 +0000
 author: Josh Gavant
+categories:
+- Announcement
+tags:
+- Event
 ---
 
 ![Kubecon Chicago 2023](/images/kubecon-chicago-2023.jpg)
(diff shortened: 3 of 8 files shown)


cjyabraham:
This is looking good. Here are a few suggestions for the design:

I think for the tags, we could borrow code and design from [the Glossary site](https://glossary.cncf.io/application-programming-interface/). See how the tags are. You can click on any of them to then [search all content for that tag](https://glossary.cncf.io/tags/?architecture=true):
Screenshot 2023-12-16 at 2 58 30 PM


For the categories, we could put it above the blog title, similar to how we do it on [the CNCF site](https://www.cncf.io/announcements/2023/11/08/cloud-native-computing-foundation-presents-the-top-end-user-award-to-spotify/):

Screenshot 2023-12-16 at 3 00 44 PM

What do you think?

abangser:
Thanks for the quick review @cjyabraham! I think your suggestions make a lot of sense.

Are these simple additions for you? I don't want to leave you with a mess, but also appreciate that you have a lot better handle on this code base than I do!

cjyabraham:
@abangser I think it should be pretty straightforward since I can lift a lot of the code from the Glossary site. I should be able to get to this in the next 3 weeks if you can wait that long?

abangser:
Absolutely. I'd rather it done right than fast and with the holidays that is actually really speedy! Thanks so much for jumping on this 🙇‍♀️

I will reach out to tag leadership to verify the intentions as well before you get started so that we have a smooth merging once done 👍👍

cjyabraham:
Ok, sounds good :)

joshgav:
Thanks for this @abangser! It looks like the tags and categories are arbitrary and anyone can choose any string, is that right? So future posts won't be constrained to an established taxonomy. That begs the question - should we establish an official tag "taxonomy" somehow? That is, so that future writers will have to choose from an established list of terms. Is that possible somehow in Hugo? If not, we can rely on reviews here, and perhaps we should document our tagging patterns.

I think the main question in this PR is if the initial set of categories and tags makes sense. To this - perhaps the two categories should be "announcement" and "article", and "community-contributed" could be a tag on some articles.

website/content/en/blog/clusters-for-all-cloud-tenants.md
```diff
@@ -1,7 +1,12 @@
 ---
 title:  "Clusters for all cloud tenants"
 date:   2022-06-02 13:04:00 +0200
-author: Josh Gavant 
+author: Josh Gavant
+categories:
+- Community
```
joshgav, line 6: Perhaps - "Article" and put "Community-contributed" as a tag?

website/content/en/blog/clusters-for-all-cloud-tenants.md
```diff
@@ -1,7 +1,12 @@
...
+tags:
+- WG Multi-tenancy
+- Technology
```
joshgav, line 9: I think you're getting at the fact that the article is technical and not some sort of announcement. Perhaps "technical" or "deep dive" would fit better?

lianmakesthings:
Whie we're on the topic of tags. 
@cjyabraham we discussed way back to introduce an "inactive" tag with which we could mark inactive Working groups that we still want to feature ont he site (e.g. operator).
Does it make sense to put that in this PR as well?

website/content/en/blog/clusters-for-all-cloud-tenants.md
abangser: I like this!

abangser:
Thanks both for the reviews! So I think a few things...

1. The "main point" of this PR is to intro the design of how a tag will show up on the page. That hasn't had a lot of pushback since Chris so nicely improved on my ideas so I suspect those are OK?
2. The next most important thing is that we have a happy set up of starting tags. I agree with what you said Josh that we have "announcement" and "article" and then we can have tag of "community contribution". Let's keep it simple and add as we go.
3. There is an idea to formalise this. I think we can add these tags to the contribution guidelines / template for blog submissions as a more "manual" tracking of this. As for automated I am not sure there is a mechanism for that so will leave it to Hugo experts!

joshgav:
@abangser agreed with all your points - once we're okay with the categories and tags I think you should merge, this is a good enhancement!

Also agree with documenting the tags in use somewhere as a "minimum viable" process. Maybe a Hugo expert will come along with a better idea eventually :smile:.

joshgav (approved):
LGTM when you're ready to merge

abangser:
Per the conversation here I am going to merge this PR with only the simple categorisation and tagging in the markdown files. This will have no visible impact for now but will be set up for testing #522 as that get completed next month!

PR #519, opened 2023-12-17
Contributions page

abangser:
We get a lot of questions about where to start with the TAG or a WG. This page is a great location for us to start supporting people with an easier path into the groups.

@techmaharaj and I put together a draft of something we could add here to get started.

### DIFF
--- a/website/content/en/contribute/_index.md
+++ b/website/content/en/contribute/_index.md
@@ -6,3 +6,25 @@ menu:
     weight: 20
 description: Ways to get involved with the TAG
 ---
+
+Here are a few suggestions on how to get started with the CNCF Technical Advisory Group (TAG) App Delivery. While you may find it useful to approach each in order, there is no requirement to do so. Find what makes sense to you and jump in!
+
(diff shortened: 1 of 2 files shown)


lianmakesthings:
Great work! Thanks for taking this on 👏

lianmakesthings (approved):

abangser:
Thank you for the rapid review @lianmakesthings !! 

I have added another commit so will wait for more reviews. I was thinking of making this commit on a new PR, but decided it is connected enough to add here. I can of course pull it out if it is more controversial.

This commit is thanks in large part to @earth2marsh and will help us continue to support new authors with more transparency and clarity on what it means to contribute!

joshgav (approved):
Thank you for this @abangser and @techmaharj!

LGTM when you're happy with it!

abangser:
Awesome! Given the double review I am inclined to merge now and update later. For example, we may extend this to include taxonomy descriptions as mentioned in #518 but I think that is likely a month out so not necessary to hold this PR up for that.

Thanks all both for the contributions and speedy reviews!

PR #534, opened 2024-01-12
Makes tag browsing of blog posts work in languages other than English

hhiroshell:
This PR makes tag browsing of blog posts work in languages other than English.

The specific changes made were as follows:
- Add ui phrase settings for tag search to the i18n/ja.toml
- Reflect the latest status of blog posts, including tags to contents in the Chinese, Korean, and Japanese websites.

PR Status:
- [x] Local test
- [x] Ready for code owner review


abangser (approved):
Thank you for doing this @hhiroshell . I am sorry that the update with tags did not work across translations and wish I had caught that sooner.

First of all, these changes LGTM both in code and in the preview env.

But also, from a strategic level, is there anything we can do to add more automation around this? This could either be a github action if it is really automatable, or if it is a process thing I would love to document it and make it more widely visible.

hhiroshell:
Hi, @abangser. Thank you for reviewing!

There are different types of content, such as blogs, whitepapers, etc., and copying procedures are not the same. Therefore, it would be expensive to fully automate and maintain the process to ensure that it works correctly.
It is also difficult to determine what to do when the original text of a translated document is updated.

What about sending a warning message to the PR when the English content is updated and the other languages have not been updated? It is not perfect, but it would be useful to prevent forgetting to update.

abangser:
> What about sending a warning message to the PR when the English content is updated and the other languages have not been updated? It is not perfect, but it would be useful to prevent forgetting to update.

Nice idea @hhiroshell. I wonder if we could put together a simple github action for that? It can be a warning rather than a failure maybe? And maybe an extra enhancement could be a tagging of people who self-nominate as a language maintainer?

I am happy to take a stab at this but it may take a moment for me to get the time so most importantly I just want to see if this is a reasonable solution as I can open an issue and see if you or anyone else can beat me to the implementation!

hhiroshell:
Thanks, @abangser! I'm happy to try making such a feature with GitHub Actions. Anyway, I agree that we should first discuss whether the feature is reasonable in an issue. I'll create it.

hhiroshell:
@abangser 
I've created [an issue](https://github.com/cncf/tag-app-delivery/issues/537), so l think we should continue discussing the GitHub Action there.

Anyway, changes in this PR should be merged to make tag browsing work. I will remind about the review in the TAG Slack channel.

abangser:
I agree @hhiroshell , I have done a review of the preview deploy and it seems to work as designed. Thank you for opening the other issue and I look forward to that one helping make all of this easier as time goes on!

@lianmakesthings / @joshgav  / @thschue I think you all have the power to approve this?

joshgav (approved):
LGTM, and brings content in other languages up to date with English content.
Thanks @hhiroshell, and thanks @abangser for the review!

PR #535, opened 2024-01-15
Installs GTM instead of just GA4

cjyabraham:
To allow us more flexibility we want to install GTM in place of just GA4. We will also be including a HubSpot tracking code using GTM. See https://github.com/cncf/cncf.io/issues/823

Also includes fix for [this build issue](https://github.com/cncf/cncf.io/issues/824)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,8 +1,2 @@
 [build]
   ignore = "exit 1"
-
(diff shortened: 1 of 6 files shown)


lianmakesthings (approved):
lgtm

Issue #485, opened 2023-10-30
Platforms Glossary v2


abangser:
Based on some feedback in the maturity model calls and reviews we are identifying a need for further investment into the Platforms glossary. This was originally a late addition to the work we did on the white paper and it is proving both useful and widely referenced so it is worth us making sure it reflects what the group is currently feeling on each of these terms.

abangser:
cc/ @joshgav @AdminTurnedDevOps and @earth2marsh

This is likely to take shape post KubeCon Chicago but I know y'all are thinking on this so want to be sure you know it is captured and top of the todos!

AdminTurnedDevOps:
Thank you! Please let me know if there's anything I can do to help out here. I'll be at KubeCon as well, so happy to connect there!

abangser:
As discussed on the last WG call, here is a google doc where we can collaborate on the glossary updates: https://docs.google.com/document/d/1qYXasDys7gSPcMcTNVY9wZVXESXbMVz5KpjaLu-WYBo/edit

joshuabezaleel:
@abangser I was thinking, should we add `Platform Orchestrator` to the glossary? I myself haven't heard about this term before until when I gave the talk with Giri on KubeCon NA 23 on our IDP, one of the attendees asked about what's the platform orchestrator we're using, which turns out it was the one that we built ourselves in-house.

And also seems like all other online resources are already mentioning this term.
- https://internaldeveloperplatform.org/platform-orchestrators/
- vendor like humanitec
- and thoughtworks (https://www.thoughtworks.com/radar/techniques/platform-orchestration) on their radar

Or this can be mentioned side-by-side with the current [Platform capability provider](https://tag-app-delivery.cncf.io/wgs/platforms/glossary/#platform-capability-providers) (?) since I think there's overlap there as well.

abangser:
Absolutely seems like a good idea! I also wonder if `FleetOps` or `Fleet Operations` which came up in a previous call may be worth adding in? We have to basically decide if we want to try and push some of the "newer" vocab or stay more reactive to what is used.

That being said, I would agree with you that platform orchestrator fits the "already being used" category anyways since it is in the TW radar!

abangser:
> @abangser I was thinking, should we add Platform Orchestrator to the glossary?

@joshuabezaleel I have taken a stab here: https://docs.google.com/document/d/1qYXasDys7gSPcMcTNVY9wZVXESXbMVz5KpjaLu-WYBo/edit

[...]

Issue #537, opened 2024-01-18
Suggestions for Introducing a GitHub Action to Improve the i18n Workflow


hhiroshell:
Currently, when adding English content (files under `website/content/en/`), we place a copy in the corresponding directories of other languages. This approach ensures that if there is no translated version available when browsing in other languages, the English content is displayed. Additionally, when updates are made, the same changes need to be applied to the corresponding files in other languages.

Our website's i18n support is relatively new, so this task is often overlooked. To prevent such oversights, I propose the introduction of a GitHub Action that would serve as a reminder (for the background to this proposal, please refer to [this PR](https://github.com/cncf/tag-app-delivery/pull/534)).

The draft idea is as follows:

- When English content is added or updated, it should be made apparent on the PR if the corresponding files in other languages have not been added or updated (for example, by displaying a warning message).
- Adding or updating content in non-English languages will not be automated. This is due to the difficulty of automatically updating in those languages when translated content is already available.

#### Request

Firstly, I would appreciate your opinion on whether this idea is reasonable. Additionally, I am open to suggestions for specific behavior of the GitHub Action. For example:

- Should we display a warning message on the PR, or should the action itself be marked as a failure?
- What about a feature that sends review requests to self-nominated language maintainers?

I would appreciate your feedback.

node:
Great suggestion ! 👍 
I meet the same case when build Chinese version website of our TAG . 
I think the way you proposed is good, we should let content in multiple languages be consistently .

antoine-sncf:
For my open source project we use https://crowdin.com/ to manage multi language translation and coordination (it detects change) and i really like the experience.
but I don't know how it could work with markdowns documents

abangser:
Thank you so much @hhiroshell for opening and so glad to see it already resonates with people. Looking at my next few weeks I think it is unlikely I can make a stab at this in the short term, but I am very happy to act as a sounding board for design/reviewer for implementations.

My initial thinking is that we can go in waves of maturity. As in, first if there is a change anywhere in the website we post a checklist of all current languages as needing to be confirmed if changes are required/made. That can get smarter maybe in that if the non-english languages are currently just english clones we could auto make that change for people. And we could potentially ask for code-reviewer type sign ups to make any change ping specific people to help with the specific languages that need it.

But honestly, just an awareness ping asking the author and the TAG reviewers to be aware is likely enough to start!

hhiroshell:
Thank you all for your input.

[...]

PR #545, opened 2024-01-30
Re-worked definitions for v2 of the platform wg glossary

earth2marsh:
Addresses #485. 

Thanks to @abangser , @joshgav , @ebcarty , and @kurktchiev for all their contributions!


abangser (changes requested):
Love this! Thanks so much for all your hard work @earth2marsh. In particular I appreciate having the clear headings for each item.

One thing we need to confirm is if/where we need to do translations for this file. Sorry I didn't think of it sooner, but I suspect this is a copy paste job into the 3 other languages we have in the site. I have set `request changes` for this review but I personally think we should add this at the end once we have approval for the English version rather than needing to change in multiple places as reviews come in 👍

earth2marsh:
That makes sense to me, thanks. I agree we should focus on closing out the English version first before investing time in translations. 

As for the question of when/if, I'm not familiar with the organization's policy here, but I hope others on this review will know! Looking now through the current version 1 of the glossary, it seems that the: 

- [Chinese version](https://github.com/cncf/tag-app-delivery/blob/main/website/content/zh/wgs/platforms/glossary/_index.md) has been translated
- [Korean version](https://github.com/cncf/tag-app-delivery/blob/main/website/content/ko/wgs/platforms/glossary/_index.md) has only headings translated
- [Japanese version](https://github.com/cncf/tag-app-delivery/blob/main/website/content/ja/wgs/platforms/glossary/_index.md) has not been translated. 

If I've understand that correctly, it suggests to me that we might publish English versions in the shorter term and update each translation as we are able over time?

abangser:
Yup, that is the list I was thinking of Marsh. I do think we can tackle the merging of translations once we settle on English, but if translations are not updated on merge, we still want to keep them sensible so I believe we have at least two options:
1. Revert back to english for the language
2. Leave the existing translation with a note suggesting checking the English page for an updated version.

This is pointing us towards needing version the glossary directory. That would allow all languages, including English, to retain  history. With versioning, my vote would be to do the 2nd option within this PR for any page with translations (Chinese and Korean) and update to English for others (Japanese).

I would then suggest that the version 2 update translations be handled in other PRs but I think it is helpful to give people a heads up and see if they have any ideas / preferences as I don't have experience translating myself.

My understanding is the following people may have an interest in the following languages:
- [Chinese version](https://github.com/cncf/tag-app-delivery/blob/main/website/content/zh/wgs/platforms/glossary/_index.md) has been translated: @node, @windsonsea, @rootsongjc, @mouuii, @mengjiao-liu, @cubxxw
- [Korean version](https://github.com/cncf/tag-app-delivery/blob/main/website/content/ko/wgs/platforms/glossary/_index.md) has only headings translated: @Eeap, @sysnet4admin, @seokho-son
- [Japanese version](https://github.com/cncf/tag-app-delivery/blob/main/website/content/ja/wgs/platforms/glossary/_index.md) has not been translated: @hhiroshell, @kaitoii11, @MasayaAoyama, @naonishijima

Just to say, This is a lot of names and I don't know how we want to handle this in the future and we have a number of things we need to tackle, ideally in an traceable way.

So in summery my suggestion is:

* Update this PR to introduce the concept of versioning to the Glossary page (@earth2marsh and I am happy to help)
* Keep the comments / suggestions in the PR focused on the content of the Glossary update in English
* Conversations about who and how to alert on translation update needs would be best suited for #537 where we are trying to find a light weight solution that works for everyone
* Translation updates to v2 of the glossary are encouraged to confirm this version is approved before tackling and then to open a new PR using the translation template (and giving feedback on if it helps or hurts the process!)

This is all just a suggestion though, and as we evolve on #537 we will learn so please do improve this workflow with us!

dstengle:
@earth2marsh Sorry for coming so late to the game, but maybe it would be possible to introduce some more of the ecosystem element to the glossary? Right now the terms point to a patform team in a more monolithic sense when in reality there may be a number of platform providers and consumers in an organization where a consumer could be an end user or another platform provider.

I think "Team" is often used flexibly to represent up to larger organizations but the distinction might be useful here and it still aligns well with other things mentioned, including team topologies.

earth2marsh:
@dstengle I think that is worth considering for v3 of the glossary. The v2 goal was to clean up and reorganize the v1 document in light of the maturity model work, and to make sure the terms defined in the glossary support the two published papers. If there is specific language you'd like to suggest, there's a Google doc linked in issue #485 that you can comment directly within. FWIW, I do think there is more we can do to highlight that multiple platforms will likely exist in any org of sufficient size, and that different people in different roles/teams may be consuming or producing or both.

joshgav (approved):
Thank you for driving this work @earth2marsh! LGTM.

roberthstrand (approved):

abangser (approved):
Given the open issues for translations, let's get this merged and we can do the improvement work to site structure soon!

Fantastic work @earth2marsh and team 👏👏👏

indyaah:
Hi folks, first of all apologies for a drive by comment from an onlooker. I have been following TAG from the sidelines for internal advocacy purposes. 

The change introduces duplicative entries for TVP, is that intentional and something that would be cleaned up in a follow up work?

joshgav:
Thank you @indyaah for catching that! I was able to quickly edit and remove the old TVP section right here in the GitHub interface. Before I merge though, could someone please double-check again that everything is good? Thanks @earth2marsh @abangser @roberthstrand !

abangser:
@joshgav I can't seem to see the new preview deploy, but absolutely can confirm the duplicate and that you're delete looks good in the diff. 

Great catch and quick fix!

earth2marsh:
Thanks @indyaah for catching that. 

@joshgav your edit LGTM. However, platform users was also duplicated, so I've removed that as well. I think this is now good to go!

joshgav:
Okay, thanks everyone! Let's merge and we'll continue to iterate together!

Issue #555, opened 2024-02-10
App Enablement Tooling img does not show in */wgs/platforms/charter.


Eeap:
<img width="787" alt="image" src="https://github.com/cncf/tag-app-delivery/assets/42088290/1fa9c116-832d-4d65-9f10-aabd3cd2a07a">

Above the image, the image url is wrong so can't get image from file. 
I think the image(`/platforms-wg/charter/img/charter_app_enablement.png`) was not copied to each content language dir. So I will add each content dir that image. :)

joshgav:
Thank you for the issue and fix @Eeap !

PR #556, opened 2024-02-10
fix: fix getting image issue wgs/platforms/charter

Eeap:
As I mentioned #555, when I translate platform WG charter.md, charter_app_enablement img is not working. I think the image is not located in each language dir, so I added that image file in each content/*/wgs/platforms/charter.


abangser (approved):
Fantastic work @Eeap! Looks great to me

joshgav (approved):
Thanks for this @Eeap!
LGTM

PR #563, opened 2024-02-24
add more speakers for tag app delivery in Paris

abangser:
Was reviewing my schedule and realised that we had some more awesome representatives from the group speaking at some of the other colo events. I fear I will have missed someone, but I tried to grab any name from people I know have contributed in the last year.

### DIFF
--- a/website/content/en/blog/kubecon-eu-2024.md
+++ b/website/content/en/blog/kubecon-eu-2024.md
@@ -45,10 +45,10 @@ We will continuously update the below schedule as more information becomes avail
 ### Tuesday: Co-located Events
 Time | Location | Topic  | Presenter
 -----|----------|--------|----
-09:00 - 19:00 | Level 7.1 Room B | [Platform Engineering Day](https://events.linuxfoundation.org/kubecon-cloudnativecon-europe/co-located-events/platform-engineering-day/)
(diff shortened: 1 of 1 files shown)


abangser:
Great question!

My logic was that there are two other talks on there that feature people involved in the TAG so I'd add anyone else who is engaged in the last 12 months with the TAG. 

I do realise now that the other two had chairs which may have been the previous bar. So if that's the case I can revert that commit and just do the coffees 👍

lianmakesthings:
Three options I see:
1. We only advertise events or talks based on title, not on the people
2. We only advertise events or talks with "official" members, i.e. chairs and TLs
3. We advertise everything we think is related to app delivery (within reason) and/or with folks working with the TAG / WGs

I don't think there's a right answer here, it's whatever we believe makes the most sense for us.
Personally, I prefer option 3. I don't anticipate it to be too overwhelming. Plus, it's nice to give people options and showcase the diversity of the community.

joshgav:
I'm okay with @lianmakesthings' option 3 if all agree, no need to make a big deal of it :D. Approval incoming.

joshgav (approved):

abangser:
Thanks both! Will merge this as is, and if anything else comes in we can always add 🙌

lianmakesthings:
Thanks for this @abangser 🚀

PR #571, opened 2024-03-04
Translate recent pages

abangser:
As I was looking at some of the translations PRs I realised that we have some content that never got ported across.

We have #537 which will hopefully make a more systematic way of supporting this, but let's get these fixed up if we can!


angellk (approved):
thanks for adding these @abangser

PR #591, opened 2024-03-26
Updating new issue template link

earth2marsh:
The old link didn't offer me the template as expected. This suggestion should address that, but please check that it works for you like it does for me!

### DIFF
--- a/website/content/en/blog/contributing-to-wg-platforms.md
+++ b/website/content/en/blog/contributing-to-wg-platforms.md
@@ -45,7 +45,7 @@ Please keep in mind this is not an exhaustive list and are very open to new idea
 
 ## Step 1 - Submission
 
-We would encourage you to generate a GitHub issue and open a Slack thread with your idea. You can use the [`community contribution` template](https://github.com/cncf/tag-app-delivery/issues/new) to create an issue and also start a thread on the [CNCF Slack](https://communityinviter.com/apps/cloud-native/cncf) in the [#wg-platforms channel](https://cloud-native.slack.com/archives/C020RHD43BP).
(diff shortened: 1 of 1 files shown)


abangser (changes requested):
This is great, thanks Marsh! Can you please just do a replace all to cover all languages? This will include Spanish as well now if you rebase off main. 

Thanks!

lianmakesthings:
@abangser with the new flow it won't be necessary to duplicate this anymore

abangser (approved):
Fantastic that the new flow unblocks this! Just to make it clear, no other language has changed this page so there is no further work to be completed.

lianmakesthings (approved):

lianmakesthings (changes requested):
The page cannot be built on the deploy preview due to this error:

> 5:46:42 PM: Error: error building site: "/opt/build/repo/website/content/en/blog/contributing-to-wg-platforms.md:1:1": "/opt/build/repo/website/layouts/_default/_markup/render-link.html:4:15": execute of template failed: template: _default/_markup/render-link.html:4:15: executing "_default/_markup/render-link.html" at <urls.Parse>: error calling Parse: parse "[https://github.com/cncf/tag-app-delivery/issues/new](https://github.com/cncf/tag-app-delivery/issues/new?assignees=&labels=proposals&projects=&template=community-contribution.yml&title=%5BCommunity%5D+%3CShort+description%3E)": first path segment in URL cannot contain colon

@earth2marsh would you mind taking a look? Does the page build correctly locally for you?

earth2marsh:
Hmmm, it looks like the URL wasn't entered properly. I don't have a local env to test, but I think my latest commit should be an improvement.

abangser (commented):
Link opens right for me on the preview deploy 👍

I did notice the same thing on the `contributing` main page. Not sure if we want to get that into this PR or later.

Also noticed that the github action for i18n shows failure, I wonder if we can make that non-blocking? Will need to investigate but I don't want to see that hold up a PR like this.

lianmakesthings (approved):

lianmakesthings:
> I did notice the same thing on the contributing main page. Not sure if we want to get that into this PR or later.

Let's do that in another PR

PR #594, opened 2024-03-26
Enable i18n with es(Spanish) for website

guillermotti:
This PR makes Spanish available in the language selection menu. It also localizes the top page and the titles of each page into Spanish for the Spanish site.

For content that has not been translated, I have copied the English content as is, following the discussion in https://github.com/cncf/tag-app-delivery/pull/492.

This is the first step to complete #593.

PR #597, opened 2024-03-27
Enable i18n with fr(French) for website

mathieu-benoit:


abangser:
Very exciting @mathieu-benoit ! Thanks for starting this branch 🎉

We actually just added Spanish and that process of adding the language first, then adding the more deep translations seems to be a great way to get started.

Your PR looks good technically, but I think we need two things:

1. A fluent french speaker's review (e.g. @antoine-sncf)
2. A copy of all English content to the french folder. This will be where translations are done, but it makes sure the site works for french even if a translation is not yet complete. 

Here is the example Spanish PR if that helps: https://github.com/cncf/tag-app-delivery/pull/594

Sheepux:
@abangser  this is a coordinated effort with Mathieu :) (we're syncing via private messages on slack)

antoine-sncf:
@abangser this is a coordinated effort with Mathieu :) (we're syncing via private messages on slack)

abangser:
Fantastic! Thanks both. If you could just make sure to clearly mark approval here once you're both agreed that would be great!

mathieu-benoit:
Thanks @abangser for the guidance, much appreciated.

I just updated the description of this PR accordingly to show the plan. Still working on it, will update this thread when ready for the reviews.

Thanks!

mathieu-benoit:
Sorry, closing this one (I messed up with the branch), this implementation is now there: https://github.com/cncf/tag-app-delivery/pull/601.

PR #598, opened 2024-03-27
Try "exit 1" in both netlify.toml files to force all builds

cjyabraham:
Builds are still being cancelled by netlifiy. It's not clear to me where the netlify.toml with the "exit 1" command should go so am trying it in both netlify.toml files, in root and in /website/.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,2 +1,2 @@
 [build]
-  ignore = "exit 1"
+ignore = "exit 1"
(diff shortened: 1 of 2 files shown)


abangser (approved):
I am no expert here, but really appreciate the attempt to make these even better. Thanks Chris!

lianmakesthings (approved):
lgtm

Issue #600, opened 2024-03-31
Platform Maturity Model website defaulting to Spanish


krumware:
It looks like the Platform Maturity Model website is now defaulting to Spanish language. Tested on multiple devices and incognito mode. This may be related to the recent Spanish translation additions in #594 

Visiting https://tag-app-delivery.cncf.io/whitepapers/platform-eng-maturity-model/ will likely show menu items in Spanish. This doesn't look to affect the Whitepaper or overhead TAG site.

guillermotti:
Hey @krumware, my apologies on this as I copied some symlinks from the English folder with website content into the Spanish folder. 

This should be solved in https://github.com/cncf/tag-app-delivery/pull/599.

The above PR should also translate the `Glossary` into Spanish. I tried locally and everything looks fine now. Double check locally if possible just in case 🙏🏽

krumware:
Tested and working. Thank you for the diligence!

PR #602, opened 2024-04-01
Removing symlinks for English folders and adding right content

guillermotti:
It solves #600.


abangser (approved):
👏 Great catch @krumware and thanks for the rapid fix @guillermotti !

lianmakesthings:
I checked this on the Deploy review and the bug still occurs when you navigate to the platforms whitepaper

lianmakesthings (approved):
lgtm

PR #604, opened 2024-04-03
Swap in PageFind instead of Google Custom Search

cjyabraham:
From https://github.com/cncf/cncf.io/issues/836:

GCS doesn't seem to produce very good search results on small sites. It'd be better to remove our reliance on Google as well. Instead we can use [PageFind](https://pagefind.app/), which is an open source local search for static sites.

It also fixes a lot of page filenames which should be `index.md` and not `_index.md`. The latter is used for lists and not individual pages.


lianmakesthings (approved):
lgtm

PR #606, opened 2024-04-04
Rename files to fix build error

cjyabraham:
I had previously changed these filenames in [this commit](https://github.com/cncf/tag-app-delivery/commit/1a9c574a726d99f8d8905cca3404f104951777d2). It didn't cause an issue for me, however, when I merged to main it started to cause the following build errors:

```
WARN  Content directory "/Users/cjyabraham/Sites/tag-app-delivery/platforms-wg/glossary" have both index.* and _index.* files, pick one.
ERROR [en] REF_NOT_FOUND: Ref "/wgs/platforms/glossary": "/Users/cjyabraham/Sites/tag-app-delivery/website/content/en/wgs/platforms/maturity-model/v1/index.md:40:28": page not found
ERROR [en] REF_NOT_FOUND: Ref "/wgs/platforms/glossary": "/Users/cjyabraham/Sites/tag-app-delivery/website/content/en/wgs/platforms/maturity-model/v1/index.md:46:225": page not found
ERROR [en] REF_NOT_FOUND: Ref "/wgs/platforms/glossary": "/Users/cjyabraham/Sites/tag-app-delivery/website/content/en/wgs/platforms/maturity-model/v1/index.md:1:1": page not found
```

I can't figure out why that is happening, however, I have reverted my change for now. The only downside of this is that the glossary files will not be indexed in the PageFind search.


lianmakesthings:
Thanks @cjyabraham I would like to take some time to investigate, but I don't have access to Netlify to retrigger deploys etc. Is there any way I could help?

lianmakesthings (approved):

cjyabraham:
I don't think you need access to Netlify. The errors are seen on my local instance. If you back out this change you should be able to reproduce them.

PR #607, opened 2024-04-04
Update website to reflect correct meeting time

lianmakesthings:
We changed the cadence of the meeting, this has not been updated on the website yet.

I made an effort to translate to Chinese, Japanese and Korean with the help of Google translate and my very limited understanding of these languages. Would appreciate a review by fluent speakers 🙏  @node @hhiroshell @Eeap

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Full charter here: <https://github.com/cncf/toc/blob/main/tags/app-delivery.md>
 
 ## Meetings
 
-1st and 3rd Wednesday of each month at 16:00 UTC ([convert to your local
+Every two weeks on Wednesday at 16:00 UTC ([convert to your local
 time](https://dateful.com/convert/utc?t=16)).
 
 Meetings are listed on the [main CNCF calendar](https://www.cncf.io/calendar/)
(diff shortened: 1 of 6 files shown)


hhiroshell (approved):
@lianmakesthings 
The Japanese translation looks fine to me!👍

Eeap:
If "Every two weeks on Wednesday" means 2nd and 4th week, it looks good to me :).
Also, it looks better word, if you change "2주마다" with "격주".

lianmakesthings:
> If "Every two weeks on Wednesday" means 2nd and 4th week, it looks good to me :)

@Eeap Unfortuntely, not necessarily. It just means in a 2-week cadence, it could be 1st and 3rd. What would be the correct translation for that?

Eeap:
> > If "Every two weeks on Wednesday" means 2nd and 4th week, it looks good to me :)
> 
> @Eeap Unfortuntely, not necessarily. It just means in a 2-week cadence, it could be 1st and 3rd. What would be the correct translation for that?

Then "2주마다" is ok, but it is not natural with following words ("2주마다 수요일" is not natural). So"격주" instead of "2주마다" is better to understand for Korean.

angellk (approved):
Lgtm

PR #601, opened 2024-04-01
Enable i18n with fr(French) for website

mathieu-benoit:
Like done with `es` here: https://github.com/cncf/tag-app-delivery/pull/594, this PR is enabling the website in `fr`.

Here are the 2 pages translated in French to review at this stage:
- `website/config.toml`
- `website/i18n/fr.toml`

The other files are a copy of `website/content/en` into `website/content/fr` as-is, without any French translation for now (this will be done later in future associated PRs). What was also done is removing these 4 symlinks and the associated content was copied manually:
- `website/content/fr/wgs/platforms/glossary/`
- `website/content/fr/wgs/platforms/maturity-model/`
- `website/content/fr/wgs/platforms/whitepaper/`
- `website/content/fr/wgs/operator/whitepaper`

Also, these following are taken into account:
- https://github.com/cncf/tag-app-delivery/commit/428df6d803c773e993eb88c63fb9398871a0f07f
- https://github.com/cncf/tag-app-delivery/pull/605

When this PR will be ready, the goal will be to have the reviews from:
- [x]  @abangser and team for consistency with other entries
- [ ]  @antoine-sncf for the French translation
- [x] @guillermotti for his recent experience with the `es` translation
- [x] The existing French translation group (who translated the CNCF glossary recently)

PR #608, opened 2024-04-04
rename maturity model folders to readd 3 languages

abangser:
As you can see below, while reviewing #601 I noticed we had lost 3 languages off the platform engineering maturity model view:
image

After investigation I noted a difference in directory name which caused this. After renaming the directories you can see that the languages reappear:
image


angellk:
@abangser it looks like French is the only one missing now when viewing netlify - could you fix and then we can merge?

abangser:
Thanks for reviewing @angellk ! French is missing as #601 is still outstanding. Once in it will be good to go 🥳

angellk (approved):
lgtm - just need to make sure French still shows up!

abangser:
That's everyone! 🎉 🇨🇳 🇰🇷 🇯🇵 🇪🇸 🇫🇷

RobertKelly:
Amazing! 👏🏻 🚀 🌏

PR #617, opened 2024-04-11
Update README.md

dstengle:
Added a resources section with link to blog tag and github issues by label.


abangser (approved):
Thanks for this David! I think we should aim to get this in ASAP. I do think we can do even more for the issue raised [in slack here](https://cloud-native.slack.com/archives/C020RHD43BP/p1712842323674459) but I never want perfection to be the enemy of progression so let's get it merged!

PR #619, opened 2024-04-13
Remove word-wrapping within paragraphs

### LINKED ISSUE #618: Website formatting around text wrapping
Hi all, we got a report in a community slack about how the whitepaper was difficult to read due to word wrapping.

Specifically:
> I found it amusing (and harder to read) when there were 3 out of 5 lines with the last word hyphenated. I used DevTools to remove the hyphens: auto styling on the p element which then left the words whole and I found it much easier to read.

Before screenshot:
image

After screenshot:
image

Is this something we have seen/heard elsewhere? @cjyabraham may have insights?

cjyabraham:
Fixes #618

### DIFF
--- a/website/assets/scss/_main.scss
+++ b/website/assets/scss/_main.scss
@@ -10,8 +10,7 @@ html .row {
 // Stops links pushing out container.
 // Experimental usage on P tag.
 // Due to Archived project github link.
-a,
-p {
+a {
 	overflow-wrap: break-word;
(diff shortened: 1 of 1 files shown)


abangser (approved):
Pulled locally and played with resizing the site. Looks like I would expect. Thanks so much for the rapid fix!

lianmakesthings (approved):

PR #621, opened 2024-04-15
Temporarily deactivate French page

lianmakesthings:
With the [new language support flow](https://github.com/cncf/tag-app-delivery/pull/612) we now require that each language not only is configured in `website/config.toml` but also has a corresponding folder in `website/content/`. Otherwise the page will remain empty.
To reduce confusion, I've disabled French in the config for now until the community has translated at least one key page so the script can do its thing and duplicate the other pages from English.
cc @abangser @mathieu-benoit

### DIFF
--- a/website/config.toml
+++ b/website/config.toml
@@ -86,12 +86,12 @@ weight = 5
 [languages.es.params]
 description = "El TAG App Delivery se centra en habilitar proyectos e iniciativas relacionados con el despliegue de aplicaciones cloud-native, incluyendo su construcción, manejo y operativa."
 
-[languages.fr]
-title = "CNCF TAG App Delivery"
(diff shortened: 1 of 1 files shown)


abangser:
This is a very exciting improvement to the repo, though I appreciate it may be confusing / frustrating for you @mathieu-benoit  and @guillermotti who helped with the original PR (#601).

Absolutely looking forward to seeing French in as soon as we can get that glossary (or whatever doc you choose) translated. Thanks again for your hard work on the original PR and looking forward to this process being easier for you moving forward!

abangser (approved):

thschue (approved):
lgtm

mathieu-benoit:
Thanks @abangser and @lianmakesthings, totally makes sense indeed, thanks for this simplification/automation around this!

JFYI: the French translation will be re-enable with the first translation of the Glossary there: https://github.com/cncf/tag-app-delivery/pull/693.

PR #622, opened 2024-04-16
Fixes whitepapers defaulting to Spanish

lianmakesthings:
the `url` property in the frontmatter of the whitepapers somehow made it so that Spanish came up as the default page (we don't know why). With the property removed from the original whitepaper, all symlinked and copied files should be fine now.


roberthstrand (approved):

angellk (approved):
lgtm!

PR #623, opened 2024-04-16
fixes Image ref in Chinese Platforms Whitepaper

lianmakesthings:
In the Chinese translation of the platforms whitepaper, a non-existing image `platforms-def.drawio.zh.png` is referenced.
This PR displays the English version of the image as long as the chinese version does not exist

@node


node:
I will check it later . Thank you.
:smile:

lianmakesthings:
Let's go with option 1, then there's no pressure on you to submit the translated image :)

abangser (approved):
Great find. Also looking forward to the translated version being stored in source control so that we have more durability with the image. 

Thanks!

node:
Created new PR #648 for adding the missing picture.

lianmakesthings:
Closed in favour of #648

PR #625, opened 2024-04-17
Set fixed urls relative to language

lianmakesthings:
Fixes https://github.com/cncf/tag-app-delivery/issues/626

Originally, we set [fixed URLs for the whitepapers](https://github.com/cncf/tag-app-delivery/pull/622). 
When these were duplicated due to the [new script for language support](https://github.com/cncf/tag-app-delivery/pull/612), they also copied the fix URL. That would lead to issues, like the default page [being hijacked by a non-default language](https://github.com/cncf/tag-app-delivery/issues/600).
By removing the fixed URL, we solved the language issue, but [broke backlinks to the whitepapers](https://github.com/cncf/tag-app-delivery/issues/626).

According to hugos docs, for multilingual support, fixed URLs need to be set [relative to the languagecode](https://gohugo.io/content-management/urls/).

This PR should bring back the fix URL, so it always points to the default language version (English) while the other languages will include their language code in the URL.


roberthstrand (approved):
Tested the deployment, and seem to be working fine. And now the URL is back to the original for English, so previous links should work again.

thschue (approved):
lgtm

Issue #633, opened 2024-04-22
[Community] Link to Platforms Definition White Paper broken in the Platform Engineering Maturity Model


tiagodolphine:
### Contribution Description

When reading the `Platform Engineering Maturity Model` it was noticed the link that references the `Platforms Definition` White Paper is not found.

For example:
https://github.com/cncf/tag-app-delivery/blob/b279406f83ead64de25572014ab421ddf26b76c7/platforms-maturity-model/v1/index.md?plain=1#L33

### Related Working Group (WG)

Platforms

### Contribution type

White paper

### Why TAG App Delivery?

This is a suggestion to fix a link that currently is not found.

lianmakesthings:
Hi @tiagodolphine, thanks for opening the issue and PR. This should be fixed in https://github.com/cncf/tag-app-delivery/pull/625

tiagodolphine:
Hi @lianmakesthings thanks for pointing out the fix.
So, all good I'll close this one.

PR #634, opened 2024-04-22
[Issue-633] Fix the link to Platforms Definition white paper in Platform Engineering Maturity Model

tiagodolphine:
Closes https://github.com/cncf/tag-app-delivery/issues/633


lianmakesthings:
Fixed already in https://github.com/cncf/tag-app-delivery/pull/625

abangser:
Thanks for the contribution @tiagodolphine, you happened to stumble in in the middle of an ongoing fix where we were prioritising keeping the URLs stable from initial release which is why we solved in a slightly different way in #625. Let us know if there are any other issues with this fix in place. Thanks!

PR #639, opened 2024-04-29
Make edit links multilingual-aware

### LINKED ISSUE #637: "Edit this page" links no longer working
The "Edit this page" links are no longer working, likely due to some of the recent restructure.

Steps to replicate:

1. Navigate to any page on the Tag App Delivery website.
2. Click "Edit this page"
3. Results in a 404 on github

image

image

cjyabraham:
Fixes #637

### DIFF
--- a/website/layouts/partials/page-meta-links.html
+++ b/website/layouts/partials/page-meta-links.html
@@ -1,8 +1,8 @@
 {{ $path := "" }}
 {{ with .File }}
-  {{ $path = .Path }}
+  {{ $path = printf "%s/%s" .Lang .Path }}
 {{ else }}
-  {{ $path = .Path }}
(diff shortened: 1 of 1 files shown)


lianmakesthings:
The problem is, that the duplicated page itself doesn't know it's duplicated. I didn't want to make it too complex, and that's why the duplication also happens during build, and not before.
I'm thinking about what the expected behaviour should be for those cases. The issue happens, when I'm on a non-default language page that has been duplicated. So if I'm clicking edit, does that mean, I want to add a translation or do I want to edit the English page? I'd argue both are valid, so it's tough to say what the fix should be.

But I think this PR will already cover a good chunk of issues so I propose to merge this one first and continue this discussion in https://github.com/cncf/tag-app-delivery/issues/637

lianmakesthings (approved):

PR #640, opened 2024-04-29
Suggested rewording and new questions

bryanrossUK:
@mluskind - here's some suggested changes after our chat last week.  I'm not wed to any, so please feel free to use or ignore as you see fit.


abangser:
Given this is a narrow scoped change, I will suggest that once there is consensus within the platform as a product project group (at least 2 or 3 👍 ) we can ping one of the TAG leads directly for approval to merge.

For future context to the tag leads, this is a WIP piece of work that we wanted version controlled, but it is not an officially "published" piece so does not need formal review.

mluskind:
@bryanrossUK @abangser we can close this PR. It was incorporated into: https://github.com/cncf/tag-app-delivery/pull/708

PR #641, opened 2024-04-29
Adding note for the maturity model

earth2marsh:
Using the `pageinfo` shortcode as discussed in Slack

### DIFF
--- a/platforms-whitepaper/latest/index.md
+++ b/platforms-whitepaper/latest/index.md
@@ -158,6 +158,10 @@ following:
    cost of deployed services through automatic instrumentation and standard
    dashboards.
 
+{{% pageinfo color="info" %}}
+Please refer to the [Platform Engineering Maturity Model](https://tag-app-delivery.cncf.io/whitepapers/platform-eng-maturity-model/) created after this paper was originally published.
+{{% /pageinfo %}}
+
 By offering consistent, compliant experiences for individual capabilities or
 sets of them, internal platforms ultimately make it easier and more efficient
 for their users to deliver valuable products.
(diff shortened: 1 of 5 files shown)


earth2marsh:
missed the v1 file in this PR. fixing...

abangser (approved):
Looks great, thanks @earth2marsh!

I have added the same change to the 3 translated pages (ja, ko, and zh). Japanese is only metadata translated, but I think we will want to call out to the KO and ZH contributors that this has been added in case they want to translate properly.

lianmakesthings:
calling @node @hhiroshell @Eeap to see if you have interest & bandwidth to translate this (can be in a separate PR)

abangser:
> can be in a separate PR

Likely _should_ be! I think there is big value in getting the right language in, but I don't want to wait on that if this is ready to go as we at least have the info there!

lianmakesthings (approved):

hhiroshell:
@lianmakesthings @abangser 

> calling @node @hhiroshell @Eeap to see if you have interest & bandwidth to translate this (can be in a separate PR)

I'll create a PR for the Japanese. 👍
But please wait until I return from GW vacation.

※ GW: https://en.wikipedia.org/wiki/Golden_Week_(Japan)

hhiroshell:
@lianmakesthings @abangser

After taking some time to think about it, I realized that the Japanese white paper has not been translated yet. 😅 
So, I'll create a PR to translate the entire paper, including the note. Please let me spend adequate time on this.

abangser:
Absolutely fair @hhiroshell, we would love to see it, but also understand the big task so do take your time and let us know if we can help in any way! Thanks 🎉

Issue #632, opened 2024-04-22
[Platforms] KubeCon EU 2024 Recap Blog Post


techmaharaj:
### Contribution Description

The idea was to recap everything that happened at KubeCon EU 2024 and make this post as the go-to post for all things platform engineering that happened at KubeCon. I've tried to cover all the aspects of the event.

Draft: https://docs.google.com/document/d/1_oc5Xy7gory3jUTku090K7JrjOaVlcQWHT4m3bc-rVc/edit?usp=sharing

Open to feedback, suggestions and comments. If there's anything that I've missed (_For eg: photos and updates from the TAG-App Delivery booth_) please suggest them in the document and I'll accept them. 

Thanks in advance :) 

_PS: I know this is delayed but I was traveling and then was caught up with work._

### Related Working Group (WG)

Platforms

### Contribution type

Blog (Default)

### Why TAG App Delivery?

Since the updates are closely related to the activities that we do as part of the Platforms Working Group, the blog post fits well here.

akelity:
LGTM :)

abangser:
This is great @techmaharaj , thanks for taking a stab at this! I can see a lot of great work happening on the google doc, what kind of timeline do you have in mind for this and what support do you need to get it over the line? Would love to see this published as it showcases how great Paris was for the group 🎉

techmaharaj:
Hi @abangser, I know a few folks have reviewed it, was hoping to review it early this week, giving time to folks to review it over the weekend. I plan to make the changes and raise a PR with the final doc tomorrow end of the day.

abangser:
That is perfect, thanks @techmaharaj !

techmaharaj:
@abangser I made all the changes and updated whatever I could. There are a couple of things on which I'm waiting for Marsh and Mathieu to revert, once that is done, we should be GTG.

abangser:
Fantastic @techmaharaj, as we spoke about in slack, please just open a PR at your convenience, and happy to work with you on this if it is helpful!

PR #647, opened 2024-05-06
Add KubeCon EU Recap Blogpost

techmaharaj:
Adding blog post for KubeCon EU Recap as mentioned in https://github.com/cncf/tag-app-delivery/issues/632


techmaharaj:
Hi, @abangser, I've addressed all the comments from the Google doc, and this is GTG from my side. Please review and do the needful.

PR #657, opened 2024-05-26
feat: add new tech lead

roberthstrand:
Thank you for the vote of confidence, and the appointment to Tech Lead. Looking forward to working even more closely with the TAG going forward :blush:


thschue (approved):
lgtm

PR #660, opened 2024-05-29
Fix Some Technical Errors in the Platform Engineering Maturity Model

hhiroshell:
This PR fixes some minor technical errors in the Platform Engineering Maturity Model that I found while working on the paper's Japanese translation.

#### The title of the Platforms White Paper
- The maturity model includes sentences referring to the White Paper, but its title is stated as "Platform Definitions White Paper."
- I fixed it to the correct title.

#### typo
- HIPPA -> HIPAA

#### A section title in the model detail
- In the model detail, the section title of `Interface` - `Level 4` is "Managed services," but in the modal table, it's "Integrated services."
- I changed the section title from "Managed services" to "Integrated services" because it seems the section's content focuses on integration around platform capabilities.

#### A link pointing to a section of the Glossary
- Now, the glossary doesn't have the section named "Platform Product Manager." The product manager is explained only in a section titled "Platform Team."
- I changed the "Product Manager" link to target the "Product Team."

### Other Changes
I've added a note explaining that the translation has not been updated from the English version to the translated one. In this case, it's only the `zh` version. I created the note as a hugo shortcode so it can be reused on other pages if needed.
- [The example of the note on preview page](https://deploy-preview-660--tag-app-delivery.netlify.app/zh/wgs/platforms/maturity-model/v1/)

### DIFF
--- a/platforms-maturity-model/v1/index.md
+++ b/platforms-maturity-model/v1/index.md
@@ -374,7 +374,7 @@ Platforms and capabilities are centrally documented and discoverable, and proces
 * A central team provides a register of available shared services across the organization.
 * Loose standards, such as requiring an automatable API and usage docs, are applied to capabilities.
 * Infrastructure as Code is used to allow easier traceability of deployed services.
-* Audits for compliance regulations such as PCI DSS or HIPPA are enabled through the service inventories.
+* Audits for compliance regulations such as PCI DSS or HIPAA are enabled through the service inventories.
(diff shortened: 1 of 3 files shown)


abangser (approved):

abangser:
Thanks so much for making these updates and for taking the care to reference the changes in the Chinese paper!

I'd love to also open an issue or at least alert the recent Chinese translators since these are fairly minor and may be something people can help with.

lianmakesthings (approved):

PR #675, opened 2024-06-05
chore: Update Youtube Link to Channel

schristoff:
Youtube Link was taking to the playlist which only has pre-April 2021 recordings. Updated the link to reflect the channel!

### DIFF
--- a/website/content/en/_index.md
+++ b/website/content/en/_index.md
@@ -39,7 +39,7 @@ as well as the [CNCF Community Calendar](https://community.cncf.io/tag-app-deliv
 
 * [Agenda and Notes](https://docs.google.com/document/d/1OykvqvhSG4AxEdmDMXilrupsX2n1qCSJUWwTc3I7AOs/edit#)
 * [Zoom Meeting](https://zoom.us/j/7276783015) (Passcode: 77777)
-* [Recordings of previous meetings](https://www.youtube.com/@cncf-tag-app-delivery)
+* [Recordings of previous meetings](https://www.youtube.com/playlist?list=PLjNzvzqUSpxJ0JfD6vrdF5bsuBaJQ2BRT)
 
 ## Leads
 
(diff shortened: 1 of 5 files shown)


website/content/en/_index.md
```diff
@@ -39,7 +39,7 @@ as well as the [CNCF Community Calendar](https://community.cncf.io/tag-app-deliv
 
 * [Agenda and Notes](https://docs.google.com/document/d/1OykvqvhSG4AxEdmDMXilrupsX2n1qCSJUWwTc3I7AOs/edit#)
 * [Zoom Meeting](https://zoom.us/j/7276783015) (Passcode: 77777)
-* [Recordings of previous meetings](https://www.youtube.com/playlist?list=PLj6h78yzYM2OHd1Ht3jiZuucWzvouAAci)
+* [Recordings of previous meetings](https://www.youtube.com/@cncf-tag-app-delivery)
```
roberthstrand, line 42: Good catch. One thing I would say is that this update should be to a playlist of actual TAG meetings, and not just the entire channel. The idea is to get to the general TAG meetings, not everything.

There is a playlist for the TAG meetings, but it needs to be updated. This is within our powers to do, while the other one was on the main CNCF channel. We have been behind on uploading meetings, but now that this is fixed, I think it's safe to point the website to the correct playlist.

The correct URL would be https://www.youtube.com/watch?v=TzE2sIZ3qeU&list=PLjNzvzqUSpxJ0JfD6vrdF5bsuBaJQ2BRT&pp=iAQB

lianmakesthings: The playlist is up to date. But i think you should use this link instead:
https://www.youtube.com/playlist?list=PLjNzvzqUSpxJ0JfD6vrdF5bsuBaJQ2BRT

schristoff: Updated to the playlist, thank y'all :)

roberthstrand (approved):
Looks great now, thank you!

PR #684, opened 2024-06-11
Add community group page to the Platforms WG info

krumware:
This adds updates to the Platforms WG page to help users find the meetings. Currently, RSVPs are limited, likely due to discoverability. I also added a suggestion for users to join the community group to receive updates.

PR #707, opened 2024-06-25
[Chore] add content from readme that is missing in website

krumware:
This content was added to the platforms-wg readme in #684, but was not also added to the website content.
This brings the two to parity.


website/content/en/wgs/platforms/_index.md
```diff
@@ -18,10 +20,12 @@ or the meetings described below.
 
 ## Meetings
 
+* Please RSVP for meetings on the [TAG App Delivery Community Groups Page](https://community.cncf.io/tag-app-delivery/)
 * Meeting schedule: 2nd and 4th Tuesday of each month at [1600 Europe/London time](https://www.timeanddate.com/worldclock/converter.html?iso=20240514T150000&p1=136)
     * [2nd Tuesday event](https://www.google.com/calendar/event?eid=cWE0ZHMyMThicDF1dDZsa3N2ZGlqYmt2Z2pfMjAyNDA1MTRUMTUwMDAwWiBsaW51eGZvdW5kYXRpb24ub3JnX281YXZqbHZ0MmNhZTlicTdhOTVlbWM0NzQwQGc&ctz=UTC)
     * [4th Tuesday event](https://www.google.com/calendar/event?eid=NGhyOHY1ZWVrbDliODY3bXU5ZnRtYWo0ZGdfMjAyNDA1MjhUMTYwMDAwWiBsaW51eGZvdW5kYXRpb24ub3JnX281YXZqbHZ0MmNhZTlicTdhOTVlbWM0NzQwQGc&ctz=UTC)
```
abangser, line 26: This is not correct as it is set to UTC, we need to update it, however [the readme one](https://github.com/cncf/tag-app-delivery/blob/main/platforms-wg/README.md?plain=1#L22) is actually set to a date in 2022. I am not sure the right way to generate a google calendar link to a recurring event 🤔

website/content/en/wgs/platforms/_index.md
```diff
@@ -18,10 +20,12 @@ or the meetings described below.
 
 ## Meetings
 
+* Please RSVP for meetings on the [TAG App Delivery Community Groups Page](https://community.cncf.io/tag-app-delivery/)
 * Meeting schedule: 2nd and 4th Tuesday of each month at [1600 Europe/London time](https://www.timeanddate.com/worldclock/converter.html?iso=20240514T150000&p1=136)
     * [2nd Tuesday event](https://www.google.com/calendar/event?eid=cWE0ZHMyMThicDF1dDZsa3N2ZGlqYmt2Z2pfMjAyNDA1MTRUMTUwMDAwWiBsaW51eGZvdW5kYXRpb24ub3JnX281YXZqbHZ0MmNhZTlicTdhOTVlbWM0NzQwQGc&ctz=UTC)
```
abangser, line 25: This is set to a date in May, I think it won't update moving forward. Plus it has UTC at the end of the URL which makes me nervous as we are now pinned to British time 👀

website/content/en/wgs/platforms/_index.md
```diff
@@ -18,10 +20,12 @@ or the meetings described below.
...
     * [4th Tuesday event](https://www.google.com/calendar/event?eid=NGhyOHY1ZWVrbDliODY3bXU5ZnRtYWo0ZGdfMjAyNDA1MjhUMTYwMDAwWiBsaW51eGZvdW5kYXRpb24ub3JnX281YXZqbHZ0MmNhZTlicTdhOTVlbWM0NzQwQGc&ctz=UTC)
     * [Full CNCF calendar](https://calendar.google.com/calendar/u/0/embed?src=linuxfoundation.org_o5avjlvt2cae9bq7a95emc4740@group.calendar.google.com)
 * Zoom: https://zoom.us/j/7276783015?pwd=R0RJMkRzQ1ZjcmE0WERGcTJTOEVyUT09
     * Passcode: 77777
 * Agendas and notes: <https://docs.google.com/document/d/1_smeS9-j-SuHJi0VXjx4g9xiD2-tgqhnlwf5oSMDQgg>
+
```
abangser, line 31:
```suggestion
* Previous meetings can be viewed on the [TAG YouTube Channel
](https://www.youtube.com/watch?v=eZYSQnsWRco&list=PLjNzvzqUSpxKH8X7wNfYZtkH_ARSeeQH0)

```

abangser (changes requested):
Thanks for the updates Colin, but I do think there are still some differences between the two pages which highlights the issue with the calendar invites too.

While I think the calendar invites being correct is blocking, I also wonder if we can auto generate one of these files from the other to remove drift which I am happy to track in a different issue/PR!

website/content/en/wgs/platforms/_index.md
krumware: I pointed out the issue with the calendar in a separate slack thread with Lian. Maybe we should just remove that link and point them to the tag app delivery events page?

lianmakesthings: Yeah that might make more sense tbh. Saves us from having to know all these instances every time we change a time.

krumware: @abangser @lianmakesthings how does it look now?

abangser (approved):
I'd love to get _something_ merged ASAP as we get people reading this and appearing at the wrong time 🙈 

I personally find RSVPing per event on the website annoying, but it is also the "official" way so I am very happy to merge this until we think of anything better!

lianmakesthings:
> I personally find RSVPing per event on the website annoying, but it is also the "official" way so I am very happy to merge this until we think of anything better!

The "anything better" option is to move to LFX, which you're free to do. If you need help with that, lmk

lianmakesthings (approved):

lianmakesthings:
@krumware there's still a DCO issue, you need to go back and sign-off two of your commits that are current unsigned (should be able to do this with `git commit --amend`)

krumware:
@lianmakesthings had some trouble with the amendment since t.here have been a number of updates since, so recreating the changes as an alternative to amending and rebasing.
#718 replaces this.

PR #711, opened 2024-07-04
Make the tab switcher of the maturity model table work in the translated version.

hhiroshell:
This PR makes the tab switcher of the model table work even if the tab names are written in multi-byte characters. Currently, this issue is occurring in the Chinese version.

And I have also made the same change in the English version. This is to ensure that there will be no issues when translations are added in the future by copying from the English version.

- [x] Local test done
- [x] Reviews by Chinese speakers

### DIFF
--- a/platforms-maturity-model/v1/index.md
+++ b/platforms-maturity-model/v1/index.md
@@ -16,7 +16,7 @@ window.onhashchange = function() {
   // get the fragment without the `#`
   const fragment = window.location.hash.substring(1)
   const found = Array.from(document.querySelectorAll('.nav-item'))
-    .filter(el => el.textContent === fragment)
+    .filter(el => el.textContent === decodeURI(fragment))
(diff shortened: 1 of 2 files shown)


abangser (commented):
> And I have also made the same change in the English version. This is to ensure that there will be no issues when translations are added in the future by copying from the English version.

This is fantastic, thank you!

And I can confirm it technically works as expected for both Chinese and English. Looking for a review for the language and then should be good to go!

hhiroshell:
@node 
Could you please check the change or involve other Chinese speakers?

node:
> @node Could you please check the change or involve other Chinese speakers?

Let me check it 😄

node (approved):
LGTM. 
This works well both in Chinese and English. Thanks for your contribution 👍

roberthstrand (approved):

PR #713, opened 2024-07-08
Update info on General Meeting in translated pages

lianmakesthings:


hhiroshell (approved):
From my point of view as a Japanese speaker, the change looks fine 👍

roberthstrand (approved):

PR #741, opened 2024-08-27
chore: use new LFX link for Platform meetings

abangser:
We have recently updated our calls to use the LFX platform which has generated a new zoom call. I have updated here to the new link.

One question is if we should be sharing this link directly or using some other LFX link in its place.


lianmakesthings:
> One question is if we should be sharing this link directly or using some other LFX link in its place.

I think this is the link we should be sharing, it doubles as zoom and "event" link

lianmakesthings (approved):

PR #754, opened 2024-09-30
Adds CNCF "hello" bar above topnav

cjyabraham:
As part of the LF push to advertise events more broadly, this deploys the "hello" bar seen at the top of the CNCF site. It will be deployed to all CNCF subsites as per [this issue](https://github.com/cncf/cncf.io/issues/598). It takes its settings from the main CNCF site and will appear only when the CNCF site has it active.

Screenshot 2024-09-30 at 2 26 12 PM

There's also a commit to improve the search experience ([more info](https://github.com/cncf/glossary/pull/3207).)

### DIFF
--- a/website/layouts/_default/search.html
+++ b/website/layouts/_default/search.html
@@ -55,7 +55,7 @@
 
 		var newUrl = baseUrl + "?" + urlParams.toString();
 		// Update the browser history (optional)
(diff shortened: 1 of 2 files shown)


lianmakesthings (approved):
LGTM

PR #793, opened 2025-01-14
Update platforms working group time converter link

ryandoyle:
Use the current date which takes into account any current daylight savings deltas.

### DIFF
--- a/website/content/en/wgs/platforms/_index.md
+++ b/website/content/en/wgs/platforms/_index.md
@@ -23,7 +23,7 @@ To register for updates and to "Join" the group, please also consider joining th
 
 ## Meetings
 
-* Meeting schedule: Every other Tuesday of each month at [1600 Europe/London time](https://www.timeanddate.com/worldclock/converter.html?iso=20250114T160000&p1=136)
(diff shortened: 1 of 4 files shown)


abangser (changes requested):
Thanks so much for noticing and taking action on this! 🚀

However I don't think this change is very sustainable. It will be out if date again in 2 months when day light savings happens again. 

I wonder if a link like this makes more sense? 

https://www.worldtimebuddy.com/?pl=1&lid=2643743,30,3451190,5391959,5128581,2147714&h=2643743&hf=1

Happy for you to propose a different one but I'm hesitant to update this page to something that will require this level of upkeep. 

In addition, we will need this to be edited for all languages as the auto comment says. You can find these pages in the same hierarchy under the website directory. 

Thanks for taking this on, it's been a niggle for a long time 🙌

ryandoyle:
Cheers @abangser, thanks for the link, much more sustainable :). I've slightly motified it to select the 4PM time slot and updated all languages

abangser (approved):
That's fantastic! Thanks for reviewing the link and extending to the other languages 🗺️ 

LGTM and will get someone from the TAG to merge (maybe @GenPage?)

PR #795, opened 2025-01-16
Adding first batch of PaaP research results

Rotfuks:
Including fullly anonymised raw transcripts and an export of the survey


krumware:
@roberthstrand We could use your feedback on this one.  I don't have a problem with the format. I think the root folder may need to be moved as part of a broader reorganization perhaps.

Regarding the content, I found several mentions of vendor solutions, but I'm not certain if those need to also be de-identified.

roberthstrand:
> @roberthstrand We could use your feedback on this one. I don't have a problem with the format. I think the root folder may need to be moved as part of a broader reorganization perhaps.

Since we do not have anywhere else to place these, for now, I think we can just get it into the repo and worry about where to move it after we have a plan.

> Regarding the content, I found several mentions of vendor solutions, but I'm not certain if those need to also be de-identified.

After looking at this for a minute, I could only spot names of public cloud vendors and the likes. Too many organizations have those in common for it to be a potential point of identification, so I think we're good there. If there is any concern about that, we need to wipe out some commits to get rid of that at this point anyway, so I'll just merge this one for now and deal with it later on :D

PR #806, opened 2025-02-07
blog composable

varodrig:
blog about composable platforms


roberthstrand:
I apologize for the delay on this one. Overall, this looks good, but I do have one request.

There are a couple of times when you refer to this blog post as a paper, a word that has a meaning in the context of the work that we do. If we can change that to something more neutral like _article_ or _blog post_, that would help us separate any official work of the TAG or the Platforms Working Group from opinionated posts on the blog.

Thank you for contributing yet again, we really appreciate it!

roberthstrand (approved):

PR #811, opened 2025-02-15
Fix url

### LINKED ISSUE #810: Link to CNCF Landscape returns 404
In https://tag-app-delivery.cncf.io/, all the links to CNCF Landscape returns404

kaitoii11:
Fixes #810 

I fixed the urls to match the original texts but would it make more sense to rewrite the paragraph to point to this link?
It seems this link would display all the projects that are under TAG app delivery.
https://landscape.cncf.io/?view-mode=card&classify=tag#tag--app-delivery


GenPage (approved):
Thank you for fixing this @kaitoii11

PR #816, opened 2025-03-11
Expand Knative description and add mention of GitOps to Outer Loop

kdubois:
@salaboy a few small additions for your consideration


danieloh30 (approved):
Thanks for the improvement @kdubois

salaboy:
@kdubois we need the DCO :)

kdubois:
@salaboy DCO :heavy_check_mark:

